PR #36, opened 2020-07-20
Fix failing action

jerelmiller:
## Description

Fixes failing action by removing the dependency on the only job in the `pr_merged` workflow.


jerelmiller:
Its because this is the only job in the workflow, which makes no sense why it would have other jobs listed as a dependency. You can see the failure here: https://github.com/newrelic/gatsby-theme-newrelic/actions/runs/175992749

> The workflow is not valid. The workflow must contain at least one job with no dependencies.

PR #42, opened 2020-07-22
Documentation for API and configuration

### LINKED ISSUE #6: Document configuration and all component APIs
Let's document configuration options and component APIs so that others looking to use the shared Gatsby theme can do so easily.

caylahamann:
Adds navigation for the README, documentation for the options and documentation for the components. Will be added to as we add more components. 

closes #6

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,6 +5,16 @@
 This repo contains Gatsby themes used by New Relic Gatsby sites, including the
 [developer](https://developer.newrelic.com) and [open source websites](https://opensource.newrelic.com).
 
+- [Installation](#installation)
+- [Getting started](#getting-started)
+- [Configuration](#configuration)
+- [Options](#options)
+- [Components](#components)
+- [Usage](#usage)
+- [Testing](#testing)
+- [Contributing](#contributing)
+- [License](#license)
+
 ## Installation
 
 ```sh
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -25,6 +35,28 @@ yarn workspace demo develop
...
+
+## Components
+
+### `Button`
+
+Easily create buttons in different styles.
+
+| Prop    | Type   | Required | Description                                                                                |
+| ------- | ------ | -------- | ------------------------------------------------------------------------------------------ |
+| variant | string | yes      | Configures the variant of the button. Must either be `'plain'`, `'primary'`, or `'normal'` |
```
jerelmiller, line 57: Should we push people to try and use the constants rather than the string values? If so, maybe we should list this as `Button.VARIANT.PLAIN`, `Button.VARIANT.PRIMARY`, and `Button.VARIANT.NORMAL`.

Thoughts?

zstix: I'm a fan of this pattern.

It usually provides a better error when there's a typo and some editors even provide autocomplete options when writing code that uses the variables. As someone who is typo-prone and lazy, these are both wins for me.

jerelmiller: Awesome. Let's document them using the constants. Thanks!

caylahamann: Okay! I'll make these changes

README.md
```diff
@@ -25,6 +35,28 @@ yarn workspace demo develop
...
+
+## Components
+
+### `Button`
+
+Easily create buttons in different styles.
+
+| Prop    | Type   | Required | Description                                                                                                                        |
+| ------- | ------ | -------- | ---------------------------------------------------------------------------------------------------------------------------------- |
+| variant | string | yes      | Configures the variant of the button. Must either be `Button.VARIANT.PLAIN`, `Button.VARIANT.PRIMARY`, and `Button.VARIANT.NORMAL` |
```
jerelmiller, line 57:
```suggestion
| variant | string | yes      | Configures the variant of the button. Must either be `Button.VARIANT.PLAIN`, `Button.VARIANT.PRIMARY`, or `Button.VARIANT.NORMAL` |
```

README.md
```diff
@@ -25,6 +35,28 @@ yarn workspace demo develop
...
+| size    | string | no       | Configures the size of the button. Can be configured to `'small'`                                                                  |
```
jerelmiller, line 58:
```suggestion
| size    | string | no       | Configures the size of the button. Can be configured to `Button.SIZE.SMALL`                                                                  |
```

PR #44, opened 2020-07-23
Add <CodeBlock />

### LINKED ISSUE #21: Add the <CodeBlock /> component
Both the developer site and the open source site have code blocks rendered on various pages. Let's add the implementation from the [developer site](https://github.com/newrelic/developer-website/blob/main/src/components/CodeBlock.js).

jerelmiller:
Closes #21 

## Description

Adds the `<CodeBlock />` and `<MDXCodeBlock />` component. This uses the implementation from the developer site, which includes line numbers, line highlighting, live previews, etc.

## Screenshots

Screen Shot 2020-07-23 at 4 03 06 PM


packages/gatsby-theme-newrelic/src/components/CodeEditor.js
```diff
@@ -0,0 +1,56 @@
...
+import PropTypes from 'prop-types';
+import { ClassNames } from '@emotion/core';
+import Editor from 'react-simple-code-editor';
+import CodeHighlight from './CodeHighlight';
+
+const CodeEditor = ({ value, language, lineNumbers, onChange }) => {
+  const lineNumberWidth = value.trim().split('\n').length.toString().length;
+
+  return (
+    <ClassNames>
```
LizBaker, line 11: ooh I didn't know this was an option with emotion, interesting

jerelmiller: Its only really needed in 1 case. The `css` API expects you to be adding it to the `className` prop to work its compile-time magic. In this case, I needed the apply the class name to a [`textareaClassName`](https://github.com/newrelic/gatsby-theme-newrelic/pull/44/files#diff-cc5dba3515fd332c99ba16b61136dd3dR32), so this utility is provided to be able to make the `css` function return a stringified classname that I can pass to other props. 

Its rare you'll need this, but good to know it exists 🙂

PR #46, opened 2020-07-23
Search component

### LINKED ISSUE #19: Add the <SearchInput /> component
We have search inputs on both sites with similar treatment. Let's add this component so we can share both the implementation and the design.

According to the [component audit](https://docs.google.com/presentation/d/19Sjd-m8fEkcrkd7U1XBSZTH94omaFgzumpd-UWwB7L4/edit#slide=id.g8b04f47957_0_104), we want to use the design from the open source site, but with the added functionality of being able to clear the input with an `x` after typing, like on the developer website.

In addition to the designs, let's consider adding a `size` prop and allow the component to render as a `large` size which will match the size on the [open source site](https://opensource.newrelic.com/explore-projects)

LizBaker:
This PR adds a search bar component 🎉 

Closes #19 

Screen Shot 2020-07-23 at 1 29 47 PM


packages/gatsby-theme-newrelic/src/components/SearchInput.js
```diff
@@ -0,0 +1,150 @@
...
+      `,
+      container: css`
+        --icon-size: 1.5rem;
+      `,
+    },
+  },
+};
+
+const StyledContainer = styled.div`
+  display: flex;
```
jerelmiller, line 112: [nit] Do we need this `display: flex` here? Since both icons are positioned absolutely, I'm not sure this is needed.

packages/gatsby-theme-newrelic/src/components/SearchInput.js
```diff
@@ -0,0 +1,150 @@
...
+  position: relative;
+  flex-direction: column;
+  width: ${(props) => (props.width ? props.width : '100%')};
```
jerelmiller, line 115: You can simplify this and get rid of the ternary by using an or.

```js
${(props) => props.width || '100%'}
```

packages/gatsby-theme-newrelic/src/components/SearchInput.js
```diff
@@ -0,0 +1,150 @@
...
+          onClick={handleClick}
+          css={css`
+            right: 1rem;
+            top: 50%;
+            transform: translateY(-50%);
+            &:hover {
+              cursor: pointer;
+            }
+          `}
+          onKeyDown={(e) => e.preventDefault()}
```
jerelmiller, line 64: [nit] It looks like you're mixing styles a bit with how you pass the function to the event handlers. I see the `onClick` passes a function declared from a variable where this one is declared inline.

I personally prefer the inline style like you have here since it means I don't need to go search for the variable to understand what the handler does. Regardless, could you try and unify the styles a bit and either go all handlers inline or all passed from variables?

packages/gatsby-theme-newrelic/src/components/SearchInput.js
```diff
@@ -0,0 +1,150 @@
...
+
+  &:focus {
+    outline: none;
+    border: 1px solid rgba(0, 126, 138, 0.6);
+    box-shadow: 0 0 0 4px rgba(0, 126, 138, 0.1);
+  }
+`;
+
+const StyledIcon = styled(Icon)`
+  align-items: center;
```
jerelmiller, line 135: Since this icon is set to `display: block`, this `align-items` property won't really do much. You can remove this.

packages/gatsby-theme-newrelic/src/components/SearchInput.js
```diff
@@ -0,0 +1,150 @@
...
+  display: block;
+  font-size: 1rem;
```
jerelmiller, line 137: Since you're setting the height of the icon in order to change its size, you shouldn't need the `font-size` property. You can safely remove this.

packages/gatsby-theme-newrelic/src/components/SearchInput.js
```diff
@@ -0,0 +1,150 @@
...
+  &:focus {
+    outline: none;
+    border: 1px solid rgba(0, 126, 138, 0.6);
+    box-shadow: 0 0 0 4px rgba(0, 126, 138, 0.1);
+  }
+`;
+
+const StyledIcon = styled(Icon)`
+  align-items: center;
+  display: block;
```
jerelmiller, line 136: I'm realizing we added this when we thought maybe its default `display` value might have been the contributing factor to the alignment of the icon. That being said, I'm not sure this is needed anymore. Could you check to see if this does anything? Could we remove this if not?

packages/gatsby-theme-newrelic/src/components/SearchInput.js
```diff
@@ -0,0 +1,150 @@
...
+    >
+      <StyledIcon
+        css={css`
+          position: absolute;
+          left: 1rem;
+          top: 50%;
+          transform: translateY(-50%);
+        `}
+        name={Icon.TYPE.SEARCH}
+        size={size}
```
jerelmiller, line 44: Since you're setting the size of the icon in the `StyledIcon` component below, you can omit this.

packages/gatsby-theme-newrelic/src/components/SearchInput.js
```diff
@@ -0,0 +1,150 @@
...
+            transform: translateY(-50%);
+            &:hover {
+              cursor: pointer;
+            }
+          `}
+          onKeyDown={(e) => e.preventDefault()}
+          type="button"
+          size={size}
+        >
+          <StyledIcon name={Icon.TYPE.X} size={size} />
```
jerelmiller, line 68: Since you're setting the size of the icon in the `StyledIcon` component below, you can omit this prop.

packages/gatsby-theme-newrelic/src/components/SearchInput.js
LizBaker: that makes ~ absolute ~ sense 🐒

packages/gatsby-theme-newrelic/src/components/SearchInput.js
LizBaker: hmm if I remove it I lose the varied sizing, though I'm not entirely sure why

jerelmiller: Oh duh, you're using that prop within `StyledIcon`. When I wrote this, somehow my brain thought you were using `Icon` directly. Ignore this! Carry on 😆

packages/gatsby-theme-newrelic/src/components/SearchInput.js
LizBaker: resolved as above!

PR #59, opened 2020-07-29
Add SwiftType search component

LizBaker:
This PR adds a search component which uses SwiftType site search and their search-ui component library

Closes [#480](https://github.com/newrelic/developer-website/issues/480), Closes [#481](https://github.com/newrelic/developer-website/issues/481), Closes [#482](https://github.com/newrelic/developer-website/issues/482), Closes [#483](https://github.com/newrelic/developer-website/issues/483),Closes [#484](https://github.com/newrelic/developer-website/issues/484)


packages/gatsby-theme-newrelic/src/components/Overlay.js
```diff
@@ -0,0 +1,112 @@
...
+              </div>
+            </animated.div>
+          )
+      )}
+    </Portal>
+  );
+};
+
+Overlay.propTypes = {
+  children: PropTypes.node,
```
zstix, line 107: We will probably want to make this required.

```suggestion
  children: PropTypes.node.isRequired,
```

packages/gatsby-theme-newrelic/src/components/Overlay.js
```diff
@@ -0,0 +1,112 @@
...
+  onClick: PropTypes.func,
```
zstix, line 108: Nitpick: `onClick` isn't very descriptive and it may be unclear to the user what this prop does. Perhaps we could rename it to something like `onCloseOverlay`? Also, we probably should make this prop required.

packages/gatsby-theme-newrelic/src/components/Portal.js
```diff
@@ -0,0 +1,16 @@
+import { useState, useEffect } from 'react';
+import { createPortal } from 'react-dom';
+
+const Portal = ({ children }) => {
```
zstix, line 4: Thinking out loud here, but perhaps we can turn this into a hook (something like `usePortal`) and expose the portal as well as some callbacks for when content gets loaded / removed from it? That way we don't need to do the `useEffect` business in the component. Not sure if that would even work.

packages/gatsby-theme-newrelic/src/components/ResultView.js
```diff
@@ -0,0 +1,128 @@
...
+
+import {
+  appendClassName,
+  getUrlSanitizer,
+} from '@elastic/react-search-ui-views/es/view-helpers';
+import { isFieldValueWrapper } from '@elastic/react-search-ui-views/es/types/FieldValueWrapper';
+import { css } from '@emotion/core';
+import { rgba } from 'polished';
+
+function getFieldType(result, field, type) {
```
zstix, line 12: Nitpick: would you be willing to update these functions to match the style of our other functions:

```suggestion
const getFieldType = (result, field, type) => {
```

packages/gatsby-theme-newrelic/src/components/ResultView.js
```diff
@@ -0,0 +1,128 @@
...
+  result,
+  onClickLink,
+  titleField,
+  urlField,
+  ...rest
+}) {
+  const fields = getEscapedFields(result);
+  const title = getEscapedField(result, titleField);
+  const url = getUrlSanitizer(URL, window.location)(getRaw(result, urlField));
+  const newRelicSite = fields.url.split('.newrelic')[0].slice(8);
```
zstix, line 68: What does this `slice(8)` do? We might want to add some comments to make this a little more clear.

packages/gatsby-theme-newrelic/src/components/ResultView.js
```diff
@@ -0,0 +1,128 @@
...
+      <div className="sui-result__header">
+        {title && !url && (
+          <h3
+            className="sui-result__title"
+            dangerouslySetInnerHTML={{ __html: title }}
+          />
+        )}
+        {title && url && (
+          <h3>
+            <a
```
zstix, line 81: We may want to leverage our `ExternalLink` component for this.

packages/gatsby-theme-newrelic/src/components/SwiftSearch.js
```diff
@@ -0,0 +1,144 @@
...
+import { css } from '@emotion/core';
+import PropTypes from 'prop-types';
+
+// TODO: move styles to emotion/wrapper
+import Icon from './Icon';
+import styles from '../styles/SwiftSearchStyles';
+
+const connector = new SiteSearchAPIConnector({
+  documentType: 'page',
+  engineKey: 'Ad9HfGjDw4GRkcmJjUut',
```
zstix, line 25: Just want to double-check, this key is okay to be public facing?

packages/gatsby-theme-newrelic/src/components/SwiftSearch.js
```diff
@@ -0,0 +1,144 @@
...
+        raw: {},
+      },
+    },
+  },
+  initialState: {
+    resultsPerPage: 10,
+  },
+};
+
+const SwiftSearch = ({ className }) => {
```
zstix, line 54: Nitpick: can we call this `SwiftTypeSearch` (or something with the full product name)? _Swift_ means something else and could get a little confusing.

packages/gatsby-theme-newrelic/src/components/SwiftSearch.js
LizBaker: yes! this is a public read-only key meant to be used in client side api calls

packages/gatsby-theme-newrelic/src/components/SwiftSearch.js
LizBaker: definitely, good point

packages/gatsby-theme-newrelic/src/components/ResultView.js
LizBaker: this is slicing off the `https://` of the url so we can just get the sub domain, can definitely add a comment in!

packages/gatsby-theme-newrelic/src/components/Overlay.js
```diff
@@ -0,0 +1,112 @@
+import React, { useEffect } from 'react';
+import PropTypes from 'prop-types';
+import { css } from '@emotion/core';
+import Icon from './Icon';
+import Portal from './Portal';
+import NewRelicLogo from './NewRelicLogo';
+import { useTransition, animated } from 'react-spring';
+
+const Overlay = ({ children, onClick, isOpen = false }) => {
```
jerelmiller, line 9: The `onClick` callback here feels a bit oddly named. If I didn't know how the guts of this component was implemented, this would make me think that I'd execute some handler just for clicking on the backdrop. I suspect this is meant to be used for the handler that should be responsible for closing the overlay. That being the case, how about naming this something like `onClose`?

packages/gatsby-theme-newrelic/src/components/ResultView.js
LizBaker: oh yeah!

packages/gatsby-theme-newrelic/index.js
```diff
@@ -6,8 +6,11 @@ export { default as HamburgerMenu } from './src/components/HamburgerMenu';
 export { default as Icon } from './src/components/Icon';
 export { default as MDXCodeBlock } from './src/components/MDXCodeBlock';
 export { default as NewRelicLogo } from './src/components/NewRelicLogo';
+export { default as Overlay } from './src/components/Overlay';
+export { default as Portal } from './src/components/Portal';
 export { default as SearchInput } from './src/components/SearchInput';
 export { default as Surface } from './src/components/Surface';
+export { default as SwiftSearch } from './src/components/SwiftSearch';
```
jerelmiller, line 13: Out of curiosity, are we thinking that sites may want to implement search outside of the global header search experience? If not, maybe we should hold off on exporting this.

packages/gatsby-theme-newrelic/src/components/SwiftSearch.js
```diff
@@ -0,0 +1,144 @@
...
+  return (
+    <div css={styles} className={className}>
+      <SearchProvider config={configOptions}>
+        <WithSearch
+          mapContextToProps={({ isLoading, results, searchTerm }) => ({
```
jerelmiller, line 59: Out of curiosity, is this not the default? Without looking at the documentation (maybe I should 😄 ), is this not the default? My guess is this gets mapped into the callback on [line 65](https://github.com/newrelic/gatsby-theme-newrelic/pull/59/files#diff-f3e8ad67873c15cac146aaeff71f04cfR65)?

Could you maybe talk through this and way this is necessary to map props back into an object of the same shape.

packages/gatsby-theme-newrelic/src/components/SwiftSearch.js
```diff
@@ -0,0 +1,144 @@
...
+                  debounceLength={500}
+                  inputView={InputView}
+                />
+                {isLoading && (
+                  <Icon
+                    css={css`
+                      margin-top: 1rem;
+                    `}
+                    size="1.5rem"
+                    name={Icon.TYPE.LOADER}
```
jerelmiller, line 81: The more I look at this loader, the more I wonder if we should bring in the loader from NR1 (not use the one from the SDK directly, but reimplement it so it looks the same). That being said, I'm not the expert or have the final design say, so my opinion might be dumb 😆 .

packages/gatsby-theme-newrelic/src/components/SwiftSearch.js
```diff
@@ -0,0 +1,144 @@
...
+              </>
+            );
+          }}
+        </WithSearch>
+      </SearchProvider>
+    </div>
+  );
+};
+
+function InputView({ getAutocomplete, getInputProps }) {
```
jerelmiller, line 109:
```suggestion
const InputView = ({ getAutocomplete, getInputProps }) => {
```

packages/gatsby-theme-newrelic/src/components/Portal.js
jerelmiller: @zstix I can explain why the `useEffect` is important here.

Portals need an existing DOM element to attach itself to so that it knows where in the DOM to transport the UI. There are a couple ways of achieving this:

1) Modify the "template" HTML to add a DOM node with a specific ID that portals can attach themselves to. This requires an existing DOM node with that `id` attribute to act as a hook for the portal to attach itself to.

2) Create a DOM node on the fly when the component mounts and use that as the root element of the portal for the duration of the Portal. Once the portal is no longer needed, remove that root element from the DOM as part of the cleanup.

We've opted for 2) in this case to make it easy to wrap a bit of UI in a portal without needing the consumer of the component to manage that lifecycle. The component itself is a really nice shortcut for someone because they don't need to manage the whole `document.createElement`/`document.getElementById` lifecycle themselves. 

Knowing that we've opted for 2) this means we need some way of handling both the creation of the root DOM node for the portal, attaching it to the existing DOM, and the removal of that element from the DOM when the portal is no longer rendered. In this case `useEffect` handles this for us. We create a DOM node using the `useState` hook, then manage when its attached/removed from the DOM using the `useEffect` hook. Without the `useEffect` hook, there isn't a super great way to manage that lifecycle (unless there is something I'm unaware of, which is also very likely 😄 )

tl;dr; This component is a thin wrapper around the `createPortal` API so that we can wrap a bit of UI with this component instead of needing others to re-implement this logic over and over just to get something to render in the UI.

Does that make sense?

LizBaker: can definitely pull it out into a custom hook
something like this?
```
const usePortal = (isOpen, onOpen, onReset) =>
  useEffect(() => {
    if (isOpen) onOpen();
    return onReset;
  }, [isOpen, onOpen, onReset]);

usePortal(
  isOpen,
  () => (document.body.style.overflow = 'hidden'),
  () => (document.body.style.overflow = null)
);

File: packages/gatsby-theme-newrelic/src/components/SwiftSearch.js (hunk 12 above)
Comment (LizBaker) [reply packages/gatsby-theme-newrelic/src/components/SwiftSearch.js:81]: I did get Daniel's okay on using this one but we can always change it!

File: packages/gatsby-theme-newrelic/src/components/SwiftSearch.js (hunk 11 above)
Comment (LizBaker) [reply packages/gatsby-theme-newrelic/src/components/SwiftSearch.js:59]: i'm not entirely sure why they're doing it this way but we were going off the example in their docs [here](https://github.com/elastic/search-ui/blob/master/ADVANCED.md#mapcontexttoprops)

File: packages/gatsby-theme-newrelic/index.js (hunk 10 above)
Comment (LizBaker) [reply packages/gatsby-theme-newrelic/index.js:13]: we weren't sure so we wanted to leave it flexible, but i'm fine with us delivering the search as part of the header

File: packages/gatsby-theme-newrelic/src/components/Portal.js (hunk 3 above)
Comment (jerelmiller) [reply packages/gatsby-theme-newrelic/src/components/Portal.js:4]: I personally find that more complicated to use. The Portal is all about transporting some bit of the React tree to another part of the DOM. In this case, you have no way of specifying what elements are being transported and passed to `createPortal`

Current API:

```js
import { Portal } from '@newrelic/gatsby-theme-newrelic'

const MyComp = () => (
  <Portal>
    // this stuff will end up transported to another part of the DOM 
  </Portal>
)
```

Proposed:

```js
import { useState } from 'react'

const MyComp = () => {
  const [isOpen, setOpen] = useState(false)

  usePortal(isOpen, () => console.log('opened'), () => console.log('reset'))

  return createPortal(???, ???)
}
```

packages/gatsby-theme-newrelic/src/components/SwiftSearch.js
jerelmiller: Oh ok! It looks like maybe a performance optimization if I'm reading that right. Thanks for pointing me there!

packages/gatsby-theme-newrelic/src/components/SwiftSearch.js
jerelmiller: If Daniel is ok with it, so am I 😄

PR #60, opened 2020-07-30
Feedback Component

zstix:
## Description
Adds a `<Feedback>` component to the theme that can be used by sites to provide feedback controls. Rather than baking in the functionality, this component leaves the messaging, alignment, and action up to the caller. Feedback can only be submitted once per page load.

Also adds two new icons and the ability for `<Button>`s to be `disabled`.

## Related Issue(s)
* Relates to [#485](https://github.com/newrelic/developer-website/issues/485)

## Screenshot(s)
**Starting state**
Screen Shot 2020-07-31 at 3 58 19 PM
**Text entered, sentiment voted state**
Screen Shot 2020-07-31 at 3 58 30 PM
**Submitted state**
Screen Shot 2020-07-31 at 3 58 34 PM


djsauble:
I like this so much. Only comment is that the stroke width on the icons is a little thin relative to the button text. Can you increase it by a pixel or so?

zstix:
I bumped the icon size up a little, which adds some thickness and makes them a little more clear (in my opinion):
Screen Shot 2020-07-30 at 3 24 48 PM

packages/gatsby-theme-newrelic/README.md
```diff
@@ -419,6 +420,36 @@ All props are forwarded to the underlying `a` tag with the exception of the
...
+```js
+import { Feedback } from '@newrelic/gatsby-theme-newrelic';
+```
+
+**Props**
+
+| Prop              | Type     | Required | Default                 | Description                                                                                                                  |
+| ----------------- | -------- | -------- | ----------------------- | ---------------------------------------------------------------------------------------------------------------------------- |
+| `align`           | enum     | no       | `Feedback.ALIGN.CENTER` | Configures the alignment of the feedback component. Must be one of `Feedback.ALIGNMENT.LEFT` or `Feedback.ALIGNMENT.CENTER`. |
+| `onPositiveClick` | function | yes      |                         | Callback function that is called once the positive button is clicked.                                                        |
```
jerelmiller, line 436:
```suggestion
| `onPositiveClick` | function | yes      |                         | Handler that is called once the positive button is clicked.                                                        |
```

The rest of the README uses the term `handler` to refer to these callback functions. This will put it more in alignment with the rest of the components.

packages/gatsby-theme-newrelic/README.md
```diff
@@ -419,6 +420,36 @@ All props are forwarded to the underlying `a` tag with the exception of the
...
+| `onNegativeClick` | function | yes      |                         | Callback function that is called once the negative button is clicked.                                                        |
```
jerelmiller, line 437:
```suggestion
| `onNegativeClick` | function | yes      |                         | Handler that is called once the negative button is clicked.                                                        |
```

The rest of the README uses the term `handler` to refer to these callback functions. This will put it more in alignment with the rest of the components.

packages/gatsby-theme-newrelic/src/components/Button.js
```diff
@@ -65,12 +69,14 @@ const Button = styled.button`
   border-style: solid;
   ${({ variant }) => styles.variant[variant]}
   ${({ size }) => styles.size[size]}
+  ${({ disabled }) => disabled === true && styles.disabled}
```
jerelmiller, line 72:
```suggestion
  ${({ disabled }) => disabled && styles.disabled}
```

Is there a reason you need to compare with `=== true` instead of just using `disabled &&` since you've specified that `disabled` is a boolean?

packages/gatsby-theme-newrelic/src/components/Feedback.js
```diff
@@ -0,0 +1,77 @@
...
+          }
+        `}
+      >
+        <Button
+          onClick={() => {
+            updateSubmitted(true);
+            onPositiveClick();
+          }}
+          variant={Button.VARIANT.PRIMARY}
+          disabled={submitted}
```
jerelmiller, line 42: Do we want users to be able to submit multiple bits of feedback? It seems a bit odd that we'd only want them to submit 1 bit of feedback, then prevent them from doing so again.

Perhaps we could use the `useTimeout` hook in combination with this state to re-enable it after a set timeout?

zstix: I wanted to gate the user feedback so that someone didn't spam on the buttons and thought that, when a user navigated to a new page the component would reset itself (was building this with the rail in mind).

packages/gatsby-theme-newrelic/src/components/Button.js
zstix: I originally had that, but it was always making it `disabled`. Not sure why and my brain wasn't working. I can take an additional look at it.

packages/gatsby-theme-newrelic/src/components/Feedback.js
jerelmiller: That makes sense. I think we should consider disabling it for like a second or 2, then re-enabling. I've been watching feedback from the launch, and there are times where the same person will add multiple feedback items for the same page as they visit. I personally would rather not prevent someone from giving us feedback just because they've already provided it for the current page. I do agree though that we should try and avoid the user spamming the buttons by clicking them a bunch of times in a row.

djsauble: As currently implemented, is it once per page, or once for the entire site?

I think it makes sense to make it once per page so people can't spam, but perhaps we could show the buttons again when the page changes since maybe the content improved (or got worse).

packages/gatsby-theme-newrelic/src/components/Button.js
jerelmiller: Huh thats odd. `false && styles.disabled` should return `false`, which would just get omitted. Maybe the `disabled `value` is set to something other than `false` in the "falsey" cases?

packages/gatsby-theme-newrelic/src/components/Feedback.js
zstix: @djsauble that was my intention behind the current functionality: navigating to a new page would "enable" them again.

That said, it might not be ideal for feedback on search / anything more dynamic.

djsauble: I was thinking we wouldn't show these feedback buttons on search. We can infer that search wasn't helpful if you don't click on anything.

demo/src/pages/index.js
```diff
@@ -215,14 +215,17 @@ const IndexPage = ({ data }) => {
...
-              alert('meh');
-            }}
-          />
+          <div
+            css={css`
+              width: 350px;
+            `}
+          >
+            <Feedback
+              onSubmit={({ sentiment, comment }) => {
```
jerelmiller, line 224: 💯 I _really_ like this API and how you've renamed the thumbs up/down as `sentiment` here. Nice work!!

PR #63, opened 2020-08-01
Add a Tag component

jerelmiller:
## Description

Create a `<Tag />` and `<TagList />` component that can be used on the various sites. Implements the design used on the [open source site](https://opensource.newrelic.com)

## Screenshots

Screen Shot 2020-07-31 at 8 40 50 PM

### DIFF
--- a/demo/src/pages/index.js
+++ b/demo/src/pages/index.js
@@ -10,6 +10,8 @@ import {
   GlobalHeader,
   SearchInput,
   Surface,
+  Tag,
+  TagList,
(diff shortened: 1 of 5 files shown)


jerelmiller:
> Is the intention to specify tags in frontmatter and then render them using this component?

Ish. The tags will be rendered using this component, but the value of the text will be driven by the URL specified in the frontmatter. That way we don't unnecessarily burden the contributor with needing to specify the right tag value.

PR #69, opened 2020-08-05
Updates URL when in overlay

caylahamann:
closes [#487](https://github.com/newrelic/developer-website/issues/487)

Updates the URL to add an overlay tag to indicate that the search is happening in the URL, also just adds it in general. This will definitely need some refactoring.


packages/gatsby-theme-newrelic/src/utils/track.js
```diff
@@ -0,0 +1,13 @@
+function track(fn, handler, before) {
```
zstix, line 1: Can we update this to use the ES6 arrow function format?

Also, we may want to include some comments documenting how this function works and what each parameter does. If you use the [JSDoc](https://jsdoc.app/tags-param.html) format, then VSCode will provide some really nice hints when utilizing the function.

packages/gatsby-theme-newrelic/src/components/GlobalHeader.js
```diff
@@ -52,6 +68,10 @@ const GlobalHeader = ({ editUrl, className, search }) => {
     siteMetadata: { repository },
   } = site;
 
+  window.history.pushState = track(window.history.pushState, () => {
```
jerelmiller, line 71: Rather than using the built-in `pushState` call, can we instead use Gatsby's `navigate` function? I'm not sure if Gatsby tracks state internally, so it would be best to try and use the Gatsby APIs as much as possible. https://www.gatsbyjs.org/docs/gatsby-link/#how-to-use-the-navigate-helper-function

jerelmiller: I stand corrected. They recommend using the `navigate` function from reach router directly when routing query params: https://www.gatsbyjs.org/docs/gatsby-link/#recommendations-for-programmatic-in-app-navigation

packages/gatsby-theme-newrelic/src/components/GlobalHeader.js
```diff
@@ -27,7 +30,20 @@ const styles = {
...
+  const [isOverlayOpen, setIsOverlayOpen] = useState(null);
+
+  const [isQueryChanged, setIsQueryChanged] = useState(false);
+
+  useEffect(() => {
+    let searchQuery = qs.parse(window.location.search);
+    if (isOverlayOpen === null && searchQuery.overlay) setIsOverlayOpen(true);
+    if (isOverlayOpen === null && !searchQuery.overlay) setIsOverlayOpen(false);
+
+    if (isOverlayOpen && window.location.search.length && !searchQuery.overlay)
```
jerelmiller, line 42: Rather than using `window.location` to access the current location, you can use the [`useLocation`](https://reach.tech/router/api/useLocation) hook from `@reach/router` directly to get the current location (Gatsby uses reach router internally).

packages/gatsby-theme-newrelic/src/utils/track.js
jerelmiller, line 1: I'm a little unsure what this method is meant for. `track` makes me think its going to track something in like google analytics, or tessen or something, but it looks like maybe this is just a result of calling multiple functions. Could you talk through the idea behind this?

packages/gatsby-theme-newrelic/src/components/GlobalHeader.js
```diff
@@ -52,6 +68,10 @@ const GlobalHeader = ({ editUrl, className, search }) => {
     siteMetadata: { repository },
   } = site;
 
+  window.history.pushState = track(window.history.pushState, () => {
+    setIsQueryChanged(!isQueryChanged);
+  });
```
jerelmiller, line 73: I notice you're not passing any dependencies to `useEffect`. That's fine, but be aware that the callback function to it will be called with every render. Is that intentional?

packages/gatsby-theme-newrelic/src/components/GlobalHeader.js
jerelmiller: Another note here, I see you're rewriting `window.history.pushState` inside of the component, which means this is going to be defined across the entire app as soon as this component is used. If this component ever unmounts (I know this is the global header, but the same principle applies), you might have a memory leak, or at least you'll confuse React because you'll be calling `setIsQueryChanged` even after the component no longer exists. ~Typically~ Always avoid overwriting globals, especially for APIs that other parts of the app depend on, and especially if it adds behavior specific to a single component. It gets hard to untangle later if you're not careful about cleaning this up.

As mentioned, the `navigate` function should help you here.

packages/gatsby-theme-newrelic/src/components/GlobalHeader.js
```diff
@@ -27,7 +30,20 @@ const styles = {
...
+  useEffect(() => {
+    let searchQuery = qs.parse(window.location.search);
+    if (isOverlayOpen === null && searchQuery.overlay) setIsOverlayOpen(true);
+    if (isOverlayOpen === null && !searchQuery.overlay) setIsOverlayOpen(false);
+
+    if (isOverlayOpen && window.location.search.length && !searchQuery.overlay)
+      setQueryStringWithoutPageReload(window.location.search + '&overlay=true');
+    if (isOverlayOpen && !window.location.search.length)
+      setQueryStringWithoutPageReload('?overlay=true');
+  }, [isOverlayOpen, isQueryChanged]);
```
jerelmiller, line 46: It doesn't look like `isQueryChanged` is used anywhere inside this effect. Is that intentional? Could you talk about what you're trying to accomplish? 

My best guess is that between this hook and the `window.history.pushState` below, you're trying to track when the location has changed so you can determine whether to open/close the overlay. If that is the case, you might consider instead using a combination of `useEffect` with the `useLocation` hook from reach router. This will play better with React because `useLocation` is meant for location change tracking.

This might look something like this:
```js
import { useLocation } from '@reach/router'
// ...
const location = useLocation()

useEffect(() => {
  // use location instead of window.location inside of here
}, [isOverlayOpen, location])
```

I believe this might satisfy what you're trying to accomplish with the `window.history.pushState` below, which also means you'll no longer need the `isQueryChanged` variable since you'll be tracking changes to the location directly.

packages/gatsby-theme-newrelic/src/components/GlobalHeader.js
```diff
@@ -80,7 +100,10 @@ const GlobalHeader = ({ editUrl, className, search }) => {
       >
         <Overlay
           isOpen={isOverlayOpen}
-          onCloseOverlay={() => setIsOverlayOpen(false)}
+          onCloseOverlay={() => {
+            setQueryStringWithoutPageReload('');
```
jerelmiller, line 104: `navigate` should also help you here :)

packages/gatsby-theme-newrelic/src/utils/track.js
jerelmiller: With the other suggestions mentioned above, you might not need this function anymore. If those other suggestions work for you, you can remove this.

packages/gatsby-theme-newrelic/src/components/GlobalHeader.js
```diff
@@ -27,7 +30,20 @@ const styles = {
 };
 
 const GlobalHeader = ({ editUrl, className, search }) => {
-  const [isOverlayOpen, setIsOverlayOpen] = useState(false);
+  const [isOverlayOpen, setIsOverlayOpen] = useState(null);
+
+  const [isQueryChanged, setIsQueryChanged] = useState(false);
+
+  useEffect(() => {
+    let searchQuery = qs.parse(window.location.search);
```
jerelmiller, line 38: [nit] If you don't want to pull in an npm package, you can also use [`URLSearchParams`](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams) which is built into the browser as a way to parse a query string. Totally fine if you want to use `qs` since it has a nice API, but just wanted you to be aware the other one exists as well :)

packages/gatsby-theme-newrelic/src/components/GlobalHeader.js
```diff
@@ -27,7 +30,20 @@ const styles = {
...
+    if (isOverlayOpen === null && searchQuery.overlay) setIsOverlayOpen(true);
+    if (isOverlayOpen === null && !searchQuery.overlay) setIsOverlayOpen(false);
+
+    if (isOverlayOpen && window.location.search.length && !searchQuery.overlay)
+      setQueryStringWithoutPageReload(window.location.search + '&overlay=true');
```
jerelmiller, line 43: I assume the `overlay=true` is meant to ensure the _search_ overlay is open when someone has searched correct? Since this is a pretty generic name, and isn't specific to search, I suggest 1 of 2 changes here. Pick your favorite.

Option 1: Rename the variable to something more specific to search (i.e. searchOverlay or something). 
Option 2: Open the overlay based on the existence of the search parameter. For example, if a search term adds a `q` query parameter to the URL, open the overlay if `q` is in the URL (even if the value is blank).

I like option 2 personally because it means you only need a single query string variable to determine whether to open the overlay instead of 2. This makes URL sharing a bit easier.

packages/gatsby-theme-newrelic/src/components/GlobalHeader.js
```diff
@@ -27,7 +30,20 @@ const styles = {
 };
 
 const GlobalHeader = ({ editUrl, className, search }) => {
-  const [isOverlayOpen, setIsOverlayOpen] = useState(false);
+  const [isOverlayOpen, setIsOverlayOpen] = useState(null);
```
jerelmiller, line 33: You may consider setting the initial state based on whether the query string parameter exists. If someone shared a link or wanted to link to the site with a pre-filled search term, we'd want to open the overlay with the search term pre-filled.

packages/gatsby-theme-newrelic/src/components/GlobalHeader.js
caylahamann: The issue with this if we use the search component outside of the overlay, which what I was thinking about. If someone uses the global header and then decides to put search in the body, then they would be taken to the overlay when they don't necessarily want to.

packages/gatsby-theme-newrelic/src/components/GlobalHeader.js
caylahamann: the problem is that when I try to check for the query string, the global header has already rendered, so I have to do it in`useEffect` in order to check for it!

packages/gatsby-theme-newrelic/src/components/GlobalHeader.js
caylahamann: Am I not passing dependencies to `useEffect`? I thought that I was when I pass the array to useEffect which has the dependencies

packages/gatsby-theme-newrelic/src/components/GlobalHeader.js
caylahamann: So what I'm trying to do is trigger useEffect when `search-ui` changes the URL so that I can append `overlay=true` to the end of the URL. So what happens is when you search is that the search component changes the URL, and since we want to use that functionality I want to figure out when they're calling pushState (which is what the `track` function was trying to accomplish).

packages/gatsby-theme-newrelic/src/utils/track.js
caylahamann: I'm going to look into using navigate and see if it solves my problem, but the current issue has to do more with the interactions between the `search-ui` changing the URL and then having us open the URL when `overlay=true` is set

packages/gatsby-theme-newrelic/src/components/GlobalHeader.js
jerelmiller: You're right! I think my brain merged this with the `useEffect` call above so I saw it as no dependencies. I see them now! Sorry about that!

packages/gatsby-theme-newrelic/src/components/GlobalHeader.js
jerelmiller: Out of curiosity, does the `useLocation()` hook work here when search ui does its thang?

caylahamann: It does not! it's because the search ui component is updating the window by `pushState` and `useLocation()` only gets the location on initial render. I can't put it inside `useEffect` because I can't put a hook inside of a callback. So I'm not sure I can utilize `useLocation` for this purpose.

packages/gatsby-theme-newrelic/src/components/GlobalHeader.js
```diff
@@ -27,7 +29,15 @@ const styles = {
 };
 
 const GlobalHeader = ({ editUrl, className, search }) => {
-  const [isOverlayOpen, setIsOverlayOpen] = useState(false);
+  const location = useLocation();
+  const [isOverlayOpen, setIsOverlayOpen] = useState(
+    Boolean(qs.parse(location.search).q)
```
jerelmiller, line 34: Does this return `true` or `false` if the query param is `q=` (no search term)? Since empty string is falsey, I suspect this will initially be set to `false`.

packages/gatsby-theme-newrelic/src/components/GlobalHeader.js
```diff
@@ -27,7 +29,15 @@ const styles = {
...
+  );
+
+  useEffect(() => {
+    if (isOverlayOpen && !qs.parse(location.search).q) navigate('/?q=');
+    if (!isOverlayOpen) navigate('/');
```
jerelmiller, line 39: Does this mean once the overlay is closed, it will always navigate back to the home page? If so, we should probably try and keep the user on the same page, just remove the query params.

packages/gatsby-theme-newrelic/src/components/SwiftTypeSearch.js
```diff
@@ -68,6 +71,11 @@ const SwiftTypeSearch = ({ className }) => {
                   searchAsYouType
                   debounceLength={500}
                   inputView={InputView}
+                  onSubmit={(searchTerm) => {
+                    let queryString = qs.parse(location.search);
```
jerelmiller, line 75:
```suggestion
                    const queryString = qs.parse(location.search);
```

packages/gatsby-theme-newrelic/src/components/SwiftTypeSearch.js
```diff
@@ -68,6 +71,11 @@ const SwiftTypeSearch = ({ className }) => {
...
+                    queryString.q = searchTerm;
+                    navigate('/' + qs.stringify(queryString));
```
jerelmiller, line 77: Does this mean that updating the search term will always route the user back to the home page?

packages/gatsby-theme-newrelic/src/components/GlobalHeader.js
```diff
@@ -31,13 +31,14 @@ const styles = {
 const GlobalHeader = ({ editUrl, className, search }) => {
   const location = useLocation();
   const [isOverlayOpen, setIsOverlayOpen] = useState(
-    Boolean(qs.parse(location.search).q)
+    Boolean(qs.parse(location.search).q?.length > 0)
```
jerelmiller, line 34: I'm sorry for soooo many comments. Perhaps I'm not seeing something obvious 😄 

Wouldn't this result in the same thing as before if `q` is set to an empty string? If my query string was `?q=`, that would mean the value of `q` is an empty string. The `q` _property_ is still present, so I want this to return `true`. 

If I'm understanding this properly, `qs.parse(location.search)` is going to result in an object that looks like the following: `{ q: '' }`. Following the above, this means that `qs.parse(location.search).q === ''` and `qs.parse(location.search).q.length === 0`. `Boolean(0)` is `false`, but we want `true` in this case because the `q` query parameter itself is present, regardless of its value. 

You might actually have an easier time with [`URLSearchParams`](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams?q=). It has a `has` method that lets you easily check for this.

```js
const query = new URLSearchParams(location.search)
query.has('q') // => true if `q` is a query param regardless of its value
```

The equivalent using `qs.parse` would be checking whether the key exists in the object:

```js
const query = qs.parse(location.search)
Object.prototype.hasOwnProperty.call(query, 'q')
```

jerelmiller:
@caylahamann I'd be happy to pair and work through this in case my comments don't make sense. Perhaps I'm having one of those days where my writing style is not easy to follow 😄

packages/gatsby-theme-newrelic/src/components/GlobalHeader.js
caylahamann: No you are so right, that was dumb on my part hahaha I changed it and pushed the work up!

packages/gatsby-theme-newrelic/src/components/GlobalHeader.js
```diff
@@ -27,7 +28,15 @@ const styles = {
...
 const GlobalHeader = ({ editUrl, className, search }) => {
-  const [isOverlayOpen, setIsOverlayOpen] = useState(false);
+  const location = useLocation();
+  const query = new URLSearchParams(location.search);
+  const [isOverlayOpen, setIsOverlayOpen] = useState(query.has('q'));
+
+  useEffect(() => {
+    if (isOverlayOpen && !new URLSearchParams(location.search).has('q'))
+      navigate(location.pathname + '?q=');
+    if (!isOverlayOpen) navigate(location.pathname);
```
jerelmiller, line 38:
```suggestion
    if (isOverlayOpen && !new URLSearchParams(location.search).has('q')) {
      navigate(location.pathname + '?q=');
    }
    
    if (!isOverlayOpen) {
      navigate(location.pathname);
    }
```

I'm finding a bit hard to parse the conditionals since at a glance it looks like the code is just indented incorrectly rather than being part of the `if` condition. Feel free to accept/ignore this suggestion, but it helps me read this a bit easier when parsing through the code.

PR #91, opened 2020-09-02
Split.io integration + Adding global header icons back

jerelmiller:
## Description

Provides the ability to configure split.io with the gatsby site from the theme. Adds back the global header GitHub actions using an A/B test from split.io.


demo/package.json
```diff
@@ -5,7 +5,7 @@
   "license": "Apache-2.0",
   "scripts": {
     "build": "gatsby build",
-    "develop": "gatsby develop -p 8001",
+    "develop": "GATSBY_NEWRELIC_ENV=development gatsby develop -p 8001",
```
caylahamann, line 8: what does this flag do?

packages/gatsby-theme-newrelic/src/components/GlobalHeader.js
```diff
@@ -15,6 +15,8 @@ import { useLocation } from '@reach/router';
 import useQueryParams from '../hooks/useQueryParams';
 import useKeyPress from '../hooks/useKeyPress';
 import { rgba } from 'polished';
+import { useTreatments, useTrack } from '@splitsoftware/splitio-react';
```
caylahamann, line 18: what do these functions do? I tried looking into the splitIO docs but I couldn't find anything

packages/gatsby-theme-newrelic/src/components/GlobalHeader.js
```diff
@@ -62,6 +65,11 @@ const GlobalHeader = ({ className, search }) => {
   `);
 
   const utmSource = UTM_SOURCES[site.siteMetadata.siteUrl];
+  const treatments = useTreatments([SPLITS.GLOBAL_HEADER_GITHUB_BUTTONS]);
```
caylahamann, line 68: hmm nevermind I think I think I understand what treatments does :)

packages/gatsby-theme-newrelic/src/utils/generateUUID.js
```diff
@@ -0,0 +1,10 @@
+const generateUUID = () => {
```
caylahamann, line 1: what is the UUID?

demo/package.json
jerelmiller: When you boot up the demo site, you'll see a warning: 

```
GATSBY_NEWRELIC_ENV env variable is not set
```

This comes from the `gatsby-plugin-newrelic` plugin. This quiets that warning 😄

packages/gatsby-theme-newrelic/src/components/GlobalHeader.js
jerelmiller: The split docs are a bit difficult to navigate. I actually had to pop open the source code to understand `useTreatments`. Here is the docs for `useTrack`: https://help.split.io/hc/en-us/articles/360038825091-React-SDK#track

Essentially this just loads the configuration for a particular "treatment" on a split and gives you the configuration for it. For example, our `developer-website_global-header-gh-buttons` split has 2 treatments: `on` and `off`. This will load whether to show the `on` or `off` treatment a particular user.

packages/gatsby-theme-newrelic/src/utils/generateUUID.js
jerelmiller: Since we don't have the concept of a logged in user, we need some other means of creating a stable user ID that persists between sessions. This function generates a [v4 UUID](https://en.wikipedia.org/wiki/Universally_unique_identifier#Version_4_(random)) that we can then use to assign as the user ID. This user ID is persisted in local storage so that we can maintain the same ID between page refreshes.

Does that help?

caylahamann: yes!

demo/package.json
caylahamann: ahhhh I see :) thank you

PR #96, opened 2020-09-03
Adds a banner component to theme

caylahamann:
adds a banner component to theme for marketing! 

Screen Shot 2020-09-03 at 2 38 54 PM
Screen Shot 2020-09-03 at 2 38 48 PM


packages/gatsby-theme-newrelic/src/components/Banner.js
```diff
@@ -0,0 +1,36 @@
+import React, { useState } from 'react';
+import { css } from '@emotion/core';
+import Button from './Button';
+import Icon from './Icon';
+
+const Banner = ({ children, color }) => {
```
jerelmiller, line 6: Could you talk me through the motivation behind making the `color` a prop? Are we planning to have multiple banners with different colors?

packages/gatsby-theme-newrelic/src/components/Banner.js
```diff
@@ -0,0 +1,36 @@
...
+  const [visible, updateVisible] = useState(true);
+
+  if (!visible) return null;
+
+  return (
+    <div
+      css={css`
+        height: 90px;
+        background-color: ${color || '--tertiary-background-color'};
```
jerelmiller, line 15: You'll need to use `var(...)` here to ensure the background color uses the CSS variable otherwise this won't work.

packages/gatsby-theme-newrelic/src/components/Banner.js
```diff
@@ -0,0 +1,36 @@
+import React, { useState } from 'react';
+import { css } from '@emotion/core';
+import Button from './Button';
+import Icon from './Icon';
+
+const Banner = ({ children, color }) => {
+  const [visible, updateVisible] = useState(true);
```
jerelmiller, line 7: Should we consider persisting this state between page refreshes? Otherwise the user will need to dismiss the banner every time they load the site which could be annoying.

If we go that route, that likely means we need to think about creating some kind of unique ID for the banner so that dismissing the banner doesn't mean the user misses out on new announcements.

packages/gatsby-theme-newrelic/src/components/Banner.js
```diff
@@ -0,0 +1,36 @@
...
+
+  if (!visible) return null;
```
jerelmiller, line 9:
```suggestion
  if (!visible) {
    return null;
  }
```

[nit] Could we use brackets around `if` statements? Maybe I'm the minority here, but I find it a bit hard to follow without the brackets. The brackets give me a super quick way to understand what is grouped inside the condition with a bit less congitive overhead. Thoughts?

packages/gatsby-theme-newrelic/src/components/Banner.js
```diff
@@ -0,0 +1,36 @@
...
+        variant={Button.VARIANT.LINK}
+        size={Button.SIZE.EXTRA_SMALL}
+        onClick={() => updateVisible(false)}
+        css={css`
+          position: absolute;
+          top: 0;
+          right: 0;
+        `}
+      >
+        <Icon size="0.875rem" name="x"></Icon>
```
jerelmiller, line 29:
```suggestion
        <Icon size="0.875rem" name="x" />
```

I think we might have a lint rule that expects this to be self-closing if there are no children.

packages/gatsby-theme-newrelic/src/components/Banner.js
```diff
@@ -0,0 +1,36 @@
...
+
+const Banner = ({ children, color }) => {
+  const [visible, updateVisible] = useState(true);
+
+  if (!visible) return null;
+
+  return (
+    <div
+      css={css`
+        height: 90px;
```
jerelmiller, line 14: Something to consider here is that this `90px` locks us into a specific size for content regardless of how much text we have inside the banner. We may also have text overflow issues for smaller screens if there is a decent amount of text inside the banner. Should we instead let the banner use its natural height? Or maybe consider making this a `min-height`?

jerelmiller (changes requested):

packages/gatsby-theme-newrelic/src/components/Banner.js
caylahamann: @jpvajda @djsauble might have some opinions about this, we were just following the acceptance criteria from the issue! linked [here](https://github.com/newrelic/developer-website/issues/615)

packages/gatsby-theme-newrelic/src/components/Banner.js
caylahamann: we wanted to make it as easy as possible for the content creators to configure or for us to configure on the website, but what are your thoughts on this?

jerelmiller: I personally am hesitant to allow this flexibility because this allows someone to pass in arbitrary color styles that may or may not fit with the theme. I would rather start out a bit rigid and set the colors ourselves so we can maintain consistency. If we find we need this flexiblity, this would be a great addition in the future.

jpvajda: A color prop is cool, but I'm thinking we may want to standardize on the color, so we don't get some weird banners being deployed. For our color palette it seems we may want to use the inverse of the header navigation color on each mode.

so if a user is on dark mode:  the banner color is: `background-color: #F4F5F5` and `font-color: #EDEEEE` and if the user is is on light mode:  the banner color is: `background-color: #22353C` and `font-color: #678082`

@djsauble  Thoughts?

packages/gatsby-theme-newrelic/src/components/Banner.js
jpvajda: Great idea!

packages/gatsby-theme-newrelic/src/components/Banner.js
jpvajda: The banner should be responsive to text input ideally. 90px is just a standard for banners, but if someone exceed thats limit the banner ideally should accommodate the test.

packages/gatsby-theme-newrelic/src/components/Banner.js
djsauble: Agree we should be opinionated about this. In particular, the screenshots in the description look like the banner can exist without a background color, which makes the header nav look like it's floating.

I'm with @jerelmiller that we should use specific colors. Let's start with his suggestions.

packages/gatsby-theme-newrelic/src/components/Banner.js
djsauble: How do images work here? If I want a full-bleed background image, is that possible? If so we might need users to be able to provide height as a prop.

jerelmiller: I think switching this to `min-height` should solve all of these scenarios (unless we want something shorter for single-line text). That way if there are taller images, it would grow with it.

PR #97, opened 2020-09-03
Add a banner to the theme!

caylahamann:
adds a banner component to theme for marketing! 

Screen Shot 2020-09-04 at 5 56 31 PM
Screen Shot 2020-09-04 at 5 56 50 PM

### DIFF
--- a/packages/gatsby-theme-newrelic/gatsby-node.js
+++ b/packages/gatsby-theme-newrelic/gatsby-node.js
@@ -27,6 +27,11 @@ exports.createSchemaCustomization = ({ actions }) => {
       contentPadding: String
       maxWidth: String
     }
+
+    type MdxFrontmatter @infer {
+      startDate: Date @dateformat(formatString: "YYYY-MM-DD")
+      endDate: Date @dateformat(formatString: "YYYY-MM-DD")
+    }
   `);
 };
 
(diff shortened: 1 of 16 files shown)


zstix:
Screen Shot 2020-09-03 at 4 46 34 PM

I cleaned up the styles a little and set up a system for us to define and use pre-built banner styles (similar to how we handle the `<Button>` component).

Some things we should still consider:
* Persisting the banner closed state (only show again when content has changed)
* Adding the ability to specify start / end times so that we don't need to push code exactly when a marketing event begins or ends

packages/gatsby-theme-newrelic/gatsby-node.js
```diff
@@ -6,11 +6,17 @@ const uniq = (arr) => [...new Set(arr)];
...
   const { program } = store.getState();
   const imagePath = path.join(program.directory, 'src/images');
+  const announcementsPath = path.join(program.directory, 'src/announcements');
 
   if (!fs.existsSync(imagePath)) {
     reporter.info('creating the images directory');
     fs.mkdirSync(imagePath, { recursive: true });
   }
+
+  if (!fs.existsSync(announcementsPath)) {
```
caylahamann, line 16: what is this checking here?

zstix: This checks if an announcements directory exists in the filesystem. If not, it will create an `announcements` directory.

zstix (approved):

PR #100, opened 2020-09-04
Add ability to specify icon defs

jerelmiller:
## Description

In some cases, we may want to render icons with a gradient for its stroke, for example with the icons on the new Nerd Days event page. This adds the ability to specify `<defs />` on the svg that can be used to define gradients.

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,6 +11,7 @@
   },
   "devDependencies": {
     "@newrelic/eslint-plugin-newrelic": "^0.3.1",
+    "@testing-library/jest-dom": "^5.11.4",
     "@testing-library/react": "^10.4.8",
     "@testing-library/react-hooks": "^3.4.1",
     "babel-jest": "^26.3.0",
--- a/setup-test-env.js
+++ b/setup-test-env.js
@@ -1,6 +1,7 @@
 /* global expect */
 import { createSerializer, matchers } from 'jest-emotion';
 import * as emotion from '@emotion/core';
+import '@testing-library/jest-dom';
 
 expect.extend(matchers);
 expect.addSnapshotSerializer(createSerializer(emotion));
--- a/setup-test-env.js
+++ b/setup-test-env.js
@@ -1,5 +1,6 @@
 /* global expect */
-import { createSerializer } from 'jest-emotion';
+import { createSerializer, matchers } from 'jest-emotion';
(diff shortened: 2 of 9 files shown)


package.json
```diff
@@ -11,6 +11,7 @@
   },
   "devDependencies": {
     "@newrelic/eslint-plugin-newrelic": "^0.3.1",
+    "@testing-library/jest-dom": "^5.11.4",
```
caylahamann, line 14: would you be able to explain what `jest-dom` is doing here?

setup-test-env.js
```diff
@@ -1,5 +1,7 @@
 /* global expect */
-import { createSerializer } from 'jest-emotion';
+import { createSerializer, matchers } from 'jest-emotion';
 import * as emotion from '@emotion/core';
+import '@testing-library/jest-dom';
 
+expect.extend(matchers);
```
caylahamann, line 6: what is `matchers` providing here?

caylahamann (commented):
I don't know enough about jest to approve this PR, but I'm going to be doing some reading of the docs and see if I can understand this a little bit better :)

package.json
jerelmiller: [`jest-dom`](https://github.com/testing-library/jest-dom) adds a bunch of really useful matchers to match against DOM-specific stuff. For example, [`.toBeVisible()`](https://github.com/testing-library/jest-dom#tobevisible) or [`.toBeInTheDocument()`](https://github.com/testing-library/jest-dom#tobeinthedocument) (which I'm using [in this PR](https://github.com/newrelic/gatsby-theme-newrelic/pull/100/files#diff-c6211204be539afd5bf8ed94af3a1f1dR46).)

Check out the documentation for the full list of matchers!

setup-test-env.js
jerelmiller: This makes all the [`jest-dom` matchers](https://github.com/testing-library/jest-dom#custom-matchers) available for use in all tests. Without this, I wouldn't be able to use matchers such as [`expect(...).toBeInTheDocument()`](https://github.com/newrelic/gatsby-theme-newrelic/pull/100/files#diff-c6211204be539afd5bf8ed94af3a1f1dR46)

caylahamann (approved):

PR #105, opened 2020-09-09
Fixes for Gatsby Build

zstix:
## Description
This PR fixes an issue where `btoa` is not available when Gatsby builds a site using this theme.

Additionally, this adds a step to our `release` workflow that attempts to build the site. This allows us to catch any issues in the theme that would break a build for a site that uses it.

### DIFF
--- a/packages/gatsby-theme-newrelic/src/components/AnnouncementBanner.js
+++ b/packages/gatsby-theme-newrelic/src/components/AnnouncementBanner.js
@@ -22,14 +22,19 @@ const findCurrentAnnouncement = (announcements) => {
   );
 };
 
-const createContentHash = (announcement) =>
-  btoa(
+const createContentHash = (announcement) => {
+  // Since bota is not available at build time, this will provide a
+  // fallback so that this function simply joins the items.
+  const btoa = global.btoa || ((arr) => arr);
+
+  return btoa(
     [
       announcement.id,
       announcement.frontmatter.startDate,
       announcement.frontmatter.endDate,
     ].join(':')
   );
+};
 
 const components = {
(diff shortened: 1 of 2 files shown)


packages/gatsby-theme-newrelic/src/components/AnnouncementBanner.js
```diff
@@ -22,14 +22,19 @@ const findCurrentAnnouncement = (announcements) => {
   );
 };
 
-const createContentHash = (announcement) =>
-  btoa(
+const createContentHash = (announcement) => {
+  // Since bota is not available at build time, this will provide a
+  // fallback so that this function simply joins the items.
+  const btoa = global.btoa || ((arr) => arr);
```
caylahamann, line 28: what is the `((arr) => arr)` doing?

zstix: This is a fallback function that simply returns whatever is passed to it. If `createContentHash` is called and `btoa` doesn't exist, we would simply return whatever was passed to the function.

In this case, we're joining the `id`, `startDate`, and `endDate` with a `:` in between. If `btoa` exists, we'll create a hash of that string. If it doesn't, we'll just return the joined string without hashing it.

jerelmiller (approved):

jerelmiller (approved):

packages/gatsby-theme-newrelic/src/components/AnnouncementBanner.js
caylahamann: okay! awesome, thank you

PR #107, opened 2020-09-16
Doesn't return banner with swifttype index when not visible

caylahamann:
I believe that this is happening because there is a swifttype index on the Banner for the annoucement banner and we're returning that, even though there is null being returned from the actual Banner. 

By adding this conditional, I think we will solve the problem and fully remove the banner and it's index


zstix (approved):
I'm not clear what Swiftype has to do with it, but it _does_ make sense to not show a banner if `visible` is set to `false`. Once this gets released, we can add it to the revert revert PR to verify if it works: https://github.com/newrelic/developer-website/pull/762

PR #126, opened 2020-10-23
fix: make callout colors more subtle

djsauble:
Before:

image

image

After:

image

image


jerelmiller (approved):
Heck YESSSSS! This looks sooooo good.

PR #131, opened 2020-10-26
Bizible tracking script

zstix:
## Description
Adds the Bizible tracking script to our websites.

## Related Issue(s)
* Relates to https://github.com/newrelic/developer-website/issues/803

### DIFF
--- a/packages/gatsby-theme-newrelic/gatsby/on-pre-render-html.js
+++ b/packages/gatsby-theme-newrelic/gatsby/on-pre-render-html.js
@@ -18,6 +18,11 @@ const onPreRenderHTML = ({ getHeadComponents, replaceHeadComponents }) => {
       rel="stylesheet"
       href="https://fonts.googleapis.com/css2?family=Ovo&display=swap"
     />,
+    <script
+      type="text/javascript"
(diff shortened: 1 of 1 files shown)


packages/gatsby-theme-newrelic/gatsby/on-pre-render-html.js
```diff
@@ -18,6 +18,11 @@ const onPreRenderHTML = ({ getHeadComponents, replaceHeadComponents }) => {
       rel="stylesheet"
       href="https://fonts.googleapis.com/css2?family=Ovo&display=swap"
     />,
+    <script
+      type="text/javascript"
+      src="//cdn.bizible.com/scripts/bizible.js"
+      async=""
```
jerelmiller, line 24: Is `async` enough without the `=""`? Or does this need to include a value?

zstix: I copied it from their [documentation](https://docs.marketo.com/display/BIZ/Adding+Bizible+Script) and I wasn't sure if simply adding `async` would turn it into `async=true` rather than an empty string. 🤷‍♂️

jerelmiller (approved):

PR #132, opened 2020-10-27
Search input in the header

jerelmiller:
Closes https://github.com/newrelic/docs-website/issues/175

## Description

Adds a search input in the header that allows a user to type a search query, then submit it to search.

## Screenshots

Screen Shot 2020-10-26 at 5 15 55 PM

### DIFF
--- a/packages/gatsby-theme-newrelic/src/components/SearchInput.js
+++ b/packages/gatsby-theme-newrelic/src/components/SearchInput.js
@@ -5,6 +5,7 @@ import styled from '@emotion/styled';
 import Icon from './Icon';
 
 const SIZES = {
+  SMALL: 'small',
   MEDIUM: 'medium',
   LARGE: 'large',
 };
@@ -76,7 +77,7 @@ SearchInput.propTypes = {
   onClear: PropTypes.func,
   value: PropTypes.string,
   width: PropTypes.string,
-  size: PropTypes.string,
+  size: PropTypes.oneOf(Object.values(SIZES)),
   style: PropTypes.object,
 };
 
@@ -86,6 +87,15 @@ export default SearchInput;
 
 const styles = {
   size: {
+    [SIZES.SMALL]: {
+      input: css`
+        font-size: 0.75rem;
+        padding: 0.25rem calc(1.5rem + var(--icon-size));
+      `,
+      container: css`
+        --icon-size: 0.75rem;
+      `,
+    },
     [SIZES.MEDIUM]: {
       input: css`
         font-size: 0.875rem;
@@ -119,6 +129,8 @@ const StyledInput = styled.input`
   border-radius: 4px;
   background: var(--primary-background-color);
   color: var(--primary-text-color);
+  transition: 0.15s ease-out;
+  line-height: 1;
   ${({ size }) => size && styles.size[size].input}
 
   &:focus {
--- a/packages/gatsby-theme-newrelic/src/components/SearchInput.js
+++ b/packages/gatsby-theme-newrelic/src/components/SearchInput.js
@@ -12,6 +12,7 @@ const SIZES = {
 
 const SearchInput = ({
   onClear,
+  onSubmit,
   value,
   width,
   size = 'medium',
@@ -24,7 +25,14 @@ const SearchInput = ({
     onClear();
   };
   const handleKeyDown = (e) => {
-    if (e.key === 'Escape' && onClear) onClear();
+    switch (e.key) {
+      case 'Escape':
+        return onClear?.();
+      case 'Enter':
+        return onSubmit?.(value);
+      default:
+      // do nothing
+    }
   };
 
   return (
@@ -65,7 +73,12 @@ const SearchInput = ({
           type="button"
           size={size}
         >
-          <StyledIcon name={Icon.TYPE.X} />
+          <StyledIcon
+            name={Icon.TYPE.X}
+            css={css`
+              display: block;
+            `}
+          />
         </StyledButton>
       )}
     </StyledContainer>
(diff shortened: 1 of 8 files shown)


roadlittledawn:
Do we want to preserve the search term in the top nav search bar when closing the overlay? Hadn't really considered the reverse of bringing the search term back to the search bar. In theory, a user will click through to a result most of the time

packages/gatsby-theme-newrelic/src/components/SearchInput.js
```diff
@@ -1,82 +1,103 @@
...
-  ...props
-}) => {
-  const handleClick = (e) => {
-    e.preventDefault();
-    onClear();
-  };
-  const handleKeyDown = (e) => {
-    if (e.key === 'Escape' && onClear) onClear();
-  };
+const SearchInput = forwardRef(
```
caylahamann, line 13: Would you be able to give me an overview of what the forwardRef is doing? I'm looking at the documentation and currently am not getting it 😅

caylahamann (commented):
Just one comment but the rest looks good to me!

packages/gatsby-theme-newrelic/src/components/SearchInput.js
jerelmiller: Yes!

So normally when you write components, passing a `ref` to it gives you access to the _instance_ of that component. For example:

```js
class MyComponent extends React.Component {
  launch() {
    alert('Launched!!')
  }

  render() {
    return <div>My component</div>
  }
}

const AnotherComponent = () => {
  const ref = useRef()

  useEffect(() => {
    ref.current.launch() // will call the `launch` function on the `MyComponent` instance
  }, [])

  return <MyComponent ref={ref} />
}
```

Notice here how I've used `ref` as "prop" passed to `MyComponent`. This is great if you actually want to interact with the _instance_ of the component and not its underlying DOM node. Let's take another example that you see a lot:

```js
const CustomInput extends React.Component {
  render() {
    return <input {...this.props} className="custom-input" />
  }
}

const AnotherComponent = () => {
  const ref = useRef()

  useEffect(() => {
    ref.current.focus()
  }, [])

  return <CustomInput ref={ref} />
}
```

This will blow up. Why? Because the _instance_ of `CustomInput` does not have a `focus` method.

Even worse is that now that functional components are the common thing, there is no way to use a `ref` with a functional component:

```js
const CustomInput = (props) => <input {...props} className="custom-input" />

const AnotherComponent = () => {
  const ref = useRef()

  useEffect(() => {
    ref.current.focus()
  }, [])

  // this will blow up because React does not allow you to add a ref on a functional component
  return <CustomInput ref={ref} /> 
}
```

So how might we fix this?

1 way would be to create a custom prop that allowed you to forward that `ref` to the underlying node. Typically you'd see this called `innerRef`:

```js
const CustomInput = ({ innerRef, ...props }) => <input ref={innerRef} {...props} className="custom-input" />

const AnotherComponent = () => {
  const ref = useRef()

  useEffect(() => {
    ref.current.focus() // works! innerRef is now attached to the underlying <input /> element
  }, [])

  return <CustomInput innerRef={ref} /> 
}
```

This works well, but notice that it forces me to learn a new concept: I need a new name for the `ref` just so that I can "forward" it to the underlying element. Now I have to be aware of which components use `ref` and which ones expose an `innerRef`.

Another way to fix this would be to rewrite the `CustomInput` to a class component and add an instance method that would forward to the element:

```js
const CustomInput extends React.Component {
  inputRef = createRef()
  
  focus() {
    this.inputRef.current.focus()
  }

  render() {
    return <input ref={this.inputRef} {...this.props} className="custom-input" />
  }
}

const AnotherComponent = () => {
  const ref = useRef()

  useEffect(() => {
    ref.current.focus() // works! ref calls the instance method `focus` on the class, which delegates to its own ref on the input
  }, [])

  return <CustomInput ref={ref} /> 
}
```

From a consumer standpoint, this is nice, because I can just use `ref` without having to worry about its implementation. I don't have to learn a new name just to forward the ref to the underlying element. On the flip-side, this sucks as a component author because I really like functional components and I want to continue using them. It would be awful to have to rewrite my components to class components, or have to create a custom prop just so that I can forward the ref to the underlying element.

This is where `forwardRef` comes in. It wraps function components so that you can forward refs to underlying elements, but still allow consumers to use the `ref` name. This is a much cleaner approach. Now the implementation looks like this:

```js
const CustomInput = forwardRef((props, ref) => <input ref={ref} {...props} className="custom-input" />)

const AnotherComponent = () => {
  const ref = useRef()

  useEffect(() => {
    ref.current.focus() // works! ref has been forwarded by the CustomInput to the underlying <input /> element 
  }, [])

  return <CustomInput ref={ref} /> 
}
```

As you can see, this is MUCH cleaner and avoids the tradeoffs you have to make above. As a consumer of the component, you only have to know 1 concept (`ref`) and as an author, you don't have to sacrifice the way you like to author components just to forward refs.

Hopefully this helps!

jerelmiller:
> Do we want to persist the search term to the top nav search bar when closing the overlay?

@roadlittledawn It felt weird to me this way, so I opted to clear out the search term when closing the overlay. Since all of the interaction happens in the overlay itself, it seems like closing the overlay is a sign of "I'm done searching". It felt a bit cleaner to clear it out right then rather than have a search term hanging out as you navigate the page.

packages/gatsby-theme-newrelic/src/components/SearchInput.js
roadlittledawn: i was wondering the same. thanks for explaining that!

caylahamann: yes! thank you so much for this explanation, that makes a lot of sense rather than passing the ref down continuously as a prop

caylahamann (approved):

PR #133, opened 2020-10-27
Filter search results to specific sites, expose option to user

roadlittledawn:
Closes [newrelic/docs-website#177](https://github.com/newrelic/docs-website/issues/177)

This adds a `filters` field to the `configOptions.searchQuery` that get passed to `<SearchProvider />` to only return results for docs, developer, and opensource sites (as set in the `type` field in swiftype) for all queries.

It also adds a `<Facet />` component to the render that allows faceting/filtering the results for the returned values in the `type` field (where we set the site values).

I didn't look into custom styling, so not sure how much control we have there. I do like that it shows how many results per site and sorts them from most to least results for a given term.

image


jerelmiller (dismissed):
Implementation looks good! Let's hold off on merging this until we've released a new version of the theme so that we can give some design attention if need be before we release this.

djsauble:
Lookin' good! I'd suggest making better use of the whitespace by arranging the checkboxes horizontally.

roadlittledawn:
Thanks, @djsauble ! I kinda dig the result count per site, personally. It helps me see what site may be best place to look for given term/concept. Plus the component orders those checkboxes by most to least results, so without those counts I could see the changing order causing some confusion. 

Will look into removing the Sites label and horizontal spacing for those checkboxes.

roadlittledawn:
Removed `SITES` label. Change filters to horizontal, tightened whitespace for page count per site.

roadlittledawn:
**re: changing the site filter to dropdown instead of checkboxes**

i think the reason those site options only allow user to filter one at a time is because of how facets work. 

when results come back with a facet specified for `type` field, it takes all the unique field values returned and enables you to filter on one. since all records are only of one `type` (one site per record), you’ll only see that one value when the results come back filtered to that one value. (example: show me only records where `type=docs` and the only value in the type field returned is then `docs` which is what populates the facet options.

it's kind of strange there isn't a component to filter, unless i'm missing something.

there is an action called `addFilter` on `<SearchProvider />` that could maybe be used?

maybe we could write a simple dropdown component to update the filter query param?

roadlittledawn:
accidentally committed code for updating left nav search input placeholder and icon to this branch :-/

roadlittledawn:
After discussion with @jerelmiller, @jpvajda , and @djsauble, updated branch in this PR to not include any filter options in search UI in favor of re-factoring search as a whole down the road. Changes for this PR only include:
* Updating input placeholder values for search inputs in header search bar and search bar in overlay
* Add a feather icon for filter
* Add optional prop to SearchInput `iconName` to support changing left nav input icon on developer site to filter icon
* Always filter search results to docs, developer, opensource site content

packages/gatsby-theme-newrelic/src/components/SwiftypeSearch.js
```diff
@@ -70,6 +77,9 @@ const SwiftypeSearch = ({ className }) => {
                   searchAsYouType
                   debounceLength={500}
                   inputView={InputView}
+                  inputProps={{
+                    placeholder: 'Search Docs, Developer, Opensource',
```
jerelmiller, line 81:
```suggestion
                    placeholder: 'Search Docs, Developer, Open Source',
```

packages/gatsby-theme-newrelic/src/icons/feather.js
```diff
@@ -21,6 +21,11 @@ export default {
       <line x1="10" y1="14" x2="21" y2="3" />
     </>
   ),
+  filter: (
+    <>
+      <polygon points="22 3 2 3 10 12.46 10 19 14 21 14 12.46 22 3" />
+    </>
```
jerelmiller, line 27:
```suggestion
    <polygon points="22 3 2 3 10 12.46 10 19 14 21 14 12.46 22 3" />
```

Since this one only has a single child, no need for the fragment around it 😄

jerelmiller (commented):
Had those 2 minor things, otherwise this looks great!

jerelmiller (approved):

PR #134, opened 2020-10-27
Footer Component

zstix:
## Description
Adds the global footer component to the theme. This was mostly taken from the [developer site repo](https://github.com/newrelic/developer-website/blob/develop/src/components/Footer.js) with some light refactoring and applying emotion.

The logo is a generic "New Relic" logo (see screenshot). Sites that use the theme can use [shadowing](https://www.gatsbyjs.com/docs/themes/shadowing) to update this logo as needed.

## Related Issue(s)
* Relates to https://github.com/newrelic/docs-website/issues/173

## Screenshot(s)
Screen Shot 2020-10-27 at 5 28 42 PM


packages/gatsby-theme-newrelic/src/components/Footer.js
```diff
@@ -0,0 +1,129 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import Icon from './Icon';
+import Logo from './Logo';
+import ExternalLink from './ExternalLink';
+import { graphql, useStaticQuery, Link } from 'gatsby';
+import { css } from '@emotion/core';
+
+const Footer = ({ fileRelativePath }) => {
```
jerelmiller, line 9: [nit] Should we call this `GlobalFooter` since we called the main header the `GlobalHeader`? It might also better signal that this isn't just a generic footer element, but should be used as the site's main footer. Thoughts?

packages/gatsby-theme-newrelic/src/components/Footer.js
```diff
@@ -0,0 +1,129 @@
...
+  const { site } = useStaticQuery(graphql`
+    query FooterQuery {
+      site {
+        siteMetadata {
+          repository
```
jerelmiller, line 14: Can we also add a hook for `defaultBranch` in the `siteMetadata` (default the value to `main`)? Its something I've been meaning to fix. In the dev site, for example, when you edit a page, it uses `main` as the base branch, which is not what we want to accept contributions from (since we branch off of `develop`). Thoughts?

packages/gatsby-theme-newrelic/src/components/Footer.js
```diff
@@ -0,0 +1,129 @@
...
+        `}
+      >
+        <Link to="/terms">
+          <Icon name="pen" size="1rem" />
+          Terms of service
+        </Link>
+
+        {fileRelativePath && (
+          <ExternalLink
+            href={`${siteMetadata.repository}/blob/main/${fileRelativePath}`}
```
jerelmiller, line 109: See above. It would be super great to use a `defaultBranch` rather than hardcoding `main` here for repos that use a `develop` branch as the base for PRs

packages/gatsby-theme-newrelic/src/components/Footer.js
```diff
@@ -0,0 +1,129 @@
...
+          >
+            <Icon name="edit" size="1rem" />
+            Edit this page
+          </ExternalLink>
+        )}
+
+        <ExternalLink href={`${siteMetadata.repository}/issues/new/choose`}>
```
jerelmiller, line 116: According to the [MMF doc](https://docs.google.com/document/d/1jzFz-hls8Q-okkmm06x7GOPAWy1PxvJL72B-vWDOULs/edit#heading=h.i047btqfbnr9), it looks like these should be outline buttons. Could you instead use `<Button as={ExternalLink} variant={Button.VARIANT.OUTLINE} />` component instead?

packages/gatsby-theme-newrelic/src/components/Footer.js
```diff
@@ -0,0 +1,129 @@
...
+            border-left: 1px dotted var(--border-color);
+
+            @media (max-width: 760px) {
+              border-left: none;
+              margin-left: 0;
+              padding: 0.5rem;
+            }
+          `}
+        >
+          Copyright &copy; 2020 New Relic Inc.
```
jerelmiller, line 75: According to the wireframe in the [MMF doc](https://docs.google.com/document/d/1jzFz-hls8Q-okkmm06x7GOPAWy1PxvJL72B-vWDOULs/edit#heading=h.i047btqfbnr9), it looks like this should be on its own row underneath the logo and the GitHub links, rather than right next to the logo. Could you update to match?

packages/gatsby-theme-newrelic/src/components/Footer.js
```diff
@@ -0,0 +1,129 @@
...
+        </div>
+      </div>
+
+      <div
```
jerelmiller, line 79: Same here. It looks like these should be in its own row with the copyright underneath the logo and GH buttons.

jerelmiller (changes requested):
Most of my feedback was just ensuring the design matched the wireframes, but otherwise this looks pretty good.

packages/gatsby-theme-newrelic/src/components/Footer.js
zstix: I went back and forth on this, since I feel like it's "global" because it's coming from the theme. To stay consistent, I'm happy to adjust this to match.

packages/gatsby-theme-newrelic/src/components/Footer.js
zstix: I like this, I'll add it to the theme and use the default of `main`.

packages/gatsby-theme-newrelic/src/components/Footer.js
jerelmiller: If we wanted to stick with the `Footer` name, would it make sense to rename `GlobalHeader` to `Header` to match?

packages/gatsby-theme-newrelic/src/components/Footer.js
zstix: Oh wow...yeah I totally missed those new designs (that screenshot wasn't present in the ticket). I'll make all the updates to get this in line with the desired look and feel.

packages/gatsby-theme-newrelic/README.md
```diff
@@ -603,6 +606,28 @@ import { Feedback } from '@newrelic/gatsby-theme-newrelic';
...
+
+_NOTE_: The logo displayed in the footer is a generic to New Relic logo, but can be changed with [shadowing](https://www.gatsbyjs.com/docs/themes/shadowing).
+
+```js
+import { GlobalFooter } from '@newrelic/gatsby-theme-newrelic';
+```
+
+**Props**
+
+| Prop               | Type    | Required | Default | Description                                                                                                   |
```
zstix, line 621: Add documentation for the `className` support.

jerelmiller (approved):

packages/gatsby-theme-newrelic/README.md
jerelmiller: Done in https://github.com/newrelic/gatsby-theme-newrelic/pull/134/commits/a0541c34052b61fead621e6f1539ef3c9e60a5b1

PR #137, opened 2020-10-28
Dont setup gatsby-plugin-mdx in theme

jerelmiller:
## Description

Trying to setup `gatsby-plugin-mdx` in the theme has proven to be more trouble than its worth. This PR removes it from the config. None of our websites are using this config, so it has been deemed safe to remove.


jerelmiller:
Turns out we rely on MDX for the announcements

PR #139, opened 2020-10-29
Make gatsby-plugin-mdx a peer dependency

jerelmiller:
## Description

Installing and configuring the `gatsby-plugin-mdx` has proven challenging and buggy. This PR no longer installs `gatsby-plugin-mdx` directly, but now requires that sites install and configure it.


roadlittledawn (approved):
Makes sense to me. I assume this gets rid of that warning message about the plugin already being defined elsewhere?

What else was buggy about it?

jerelmiller:
>  I assume this gets rid of that warning message about the plugin already being defined elsewhere

Not quite. That still happens.

After trying to upgrade the gatsby theme on the open source site, there is something wonky about how the mdx plugin is configured. It doesn't seem to boot unless I move the MDX config to the gatsby theme configuration, which has its own weirdness.

I figured best to make this a peer dependency for now and let the individual sites have maintain their own mdx config.

PR #143, opened 2020-10-30
CookieConsentDialog component, usePrevious hook, gdprTracking configuration

caylahamann:
# Description 

This PR: 
- Adds a CookieConsentDialog component to the theme
- Adds the usePrevious hook from the dev site into the theme 
- Adds gdpr tracking configuration for the theme


packages/gatsby-theme-newrelic/README.md
```diff
@@ -1329,6 +1368,41 @@ const MyComponent = () => {
...
+  const [count, setCount] = useState(0);
+  const prevCount = usePrevious(count);
+
+  return (
+    <div>
+      <span>{`current count is ${count}, previous count is ${prevCount}`}
+      <button onClick ={()=>setCount(count+1)}>
+    </div>
+  )
+
```
jerelmiller, line 1401:
```suggestion
```

packages/gatsby-theme-newrelic/README.md
```diff
@@ -1329,6 +1368,41 @@ const MyComponent = () => {
 };
 ```
 
+### `usePrevious`
+
+A hook that gets the previous state of a stateful value. Useful inorder to do comparing if the state has changed.
```
jerelmiller, line 1373:
```suggestion
A hook that gets the previous state of a stateful value. Useful to compare if the state has changed between render cycles.
```

packages/gatsby-theme-newrelic/README.md
```diff
@@ -1329,6 +1368,41 @@ const MyComponent = () => {
...
+
+```js
+import { usePrevious } from '@newrelic/gatsby-theme-newrelic';
+```
+
+**Arguments**
+
+- `value` _(any)_: Value to track its' previous state.
```
jerelmiller, line 1381:
```suggestion
- `value` _(any)_: Value to track it's previous state.
```

packages/gatsby-theme-newrelic/README.md
```diff
@@ -1329,6 +1368,41 @@ const MyComponent = () => {
...
+
+**Returns**
+
+`any` - The previous value to be computed.
```
jerelmiller, line 1385:
```suggestion
`any` - The previous value.
```

packages/gatsby-theme-newrelic/README.md
```diff
@@ -1329,6 +1368,41 @@ const MyComponent = () => {
...
+
+**Examples**
+
+```js
+
```
jerelmiller, line 1390:
```suggestion
```

packages/gatsby-theme-newrelic/README.md
```diff
@@ -1329,6 +1368,41 @@ const MyComponent = () => {
...
+const myComponent = () => {
+  const [count, setCount] = useState(0);
+  const prevCount = usePrevious(count);
+
+  return (
+    <div>
+      <span>{`current count is ${count}, previous count is ${prevCount}`}
+      <button onClick ={()=>setCount(count+1)}>
```
jerelmiller, line 1398:
```suggestion
      <button onClick ={()=>setCount(count + 1)}>
```

packages/gatsby-theme-newrelic/README.md
```diff
@@ -1329,6 +1368,41 @@ const MyComponent = () => {
...
+    </div>
+  )
+
+}
+
```
jerelmiller, line 1403:
```suggestion
```

demo/package.json
```diff
@@ -14,6 +14,7 @@
     "@newrelic/gatsby-theme-newrelic": "^1.1.10",
     "gatsby": "^2.24.80",
     "gatsby-plugin-mdx": "^1.2.52",
+    "js-cookie": "^2.2.1",
```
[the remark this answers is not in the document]
jerelmiller: Ah, I see it installed below. Since this is a dependency of the theme, you don't need to install it elsewhere.

```suggestion
```

packages/gatsby-theme-newrelic/src/components/CookieConsentDialog.js
```diff
@@ -0,0 +1,149 @@
+import React, { useState } from 'react';
+import PropTypes from 'prop-types';
+import Cookies from 'js-cookie';
+import { css } from '@emotion/core';
+import ExternalLink from './ExternalLink';
+import Button from './Button';
+
+const CookieConsentDialog = ({
+  cookieName,
```
jerelmiller, line 9: Since we are moving this to the theme, I think we should really be opinionated for this cookie dialog since it is very specific to New Relic. That means not leaving the cookie name up to consumers of this component, but rather we will set it for them.

I really like that we are pulling the cookie consent into the theme (its loongggg overdue). In doing so, I think we should strive for as minimal configuration as possible. This is one of those values that is not really going to need to be tweaked site-to-site.

packages/gatsby-theme-newrelic/gatsby-config.js
```diff
@@ -32,6 +32,10 @@ module.exports = ({ newrelic, robots = {} }) => {
           path: 'src/announcements',
         },
       },
+      gdprTracking && {
```
jerelmiller, line 35: This is another area where I think we can be very opinionated. We know our sites are going to use google analytics. I think the key is also going to be the same. Rather than allowing the entire plugin to be configured, I think we should configure it as much as possible up-front in the theme and only expose the areas that need customization (like the GA tracking ID):

Here's what I'd propose for the configuration:

```
// developer-site/gatsby-config.js
{
  resolve: '@newrelic/gatsby-theme-newrelic',
  options: {
    // ...
    gaTrackingId: 'UA-1234'
  }
}
```

This allows us to enforce consistency across our sites without putting that burden back on the consumers of the theme. We know we want to lazy load the cookies, we know we are going to use google analytics, etc.

demo/src/pages/index.js
```diff
@@ -38,10 +41,28 @@ const liveCodeSample = `
...
+
+  const previousCookieConsent = usePrevious(cookieConsent);
+
+  const handleCookieConsent = (answer) => {
+    Cookies.set(gdprConsentCookieName, String(!!answer), options);
+    answer && setCookieConsent(true);
+  } 
+
+  useEffect(()=> {
+      if (cookieConsent && previousCookieConsent === false) {
```
jerelmiller, line 62: Can you run prettier on this file?

I wonder if your editor has turned prettier off or something. I've noticed a couple files that seem to be mis-styled. Could you double check to make sure prettier is formatting your files properly?

demo/src/pages/index.js
```diff
@@ -38,10 +41,28 @@ const liveCodeSample = `
...
+        alert(`${gdprConsentCookieName} has been approved`)
+      } 
+  }, [cookieConsent, previousCookieConsent])
```
jerelmiller, line 65: You'll notice this is a TON of boilerplate that consumers of the theme would need to concern themselves with. By moving it to the theme, I think we should be pretty opinionated and try and take as much of the pain out of implementing cookie consent as much as possible. To me, that would be: I import the component, render it, and thats it. I don't need to concern myself with setting cookies, dealing with tracking, etc.

I've posted some suggestions on how to achieve this. Ideally this demo file would look like this:

```js
import { CookieConsentDialog } from '@newrelic/gatsby-theme-newrelic'

const HomePage = () => {
  return (
    // ...
    <CookieConstentDialog />
  )
}
```

packages/gatsby-theme-newrelic/src/components/CookieConsentDialog.js
```diff
@@ -0,0 +1,149 @@
...
+  handleCookieConsent
```
jerelmiller, line 10: Do we need to expose a handler here for when the cookie consent is accepted or denied? I think this component should do as much of the heavy lifting as possible. That would include: reading from the cookie for the initial state, setting the cookie when the user chooses to accept/deny the terms, and initiating the tracking code if the user decides to accept the cookie consent. To me, this is one of those components that makes sense to be very opinionated because the cookie consent dialog itself is very opinionated to New Relic sites (regular Joe is not going to be installing this theme and using this component out-of-the-box for example).

I think for now, it makes sense to remove this handler and add it back when we have a use case for it. Until then, lets try and do a lot of the heavy lifting here.

jerelmiller (changes requested):

packages/gatsby-theme-newrelic/src/components/CookieConsentDialog.js
caylahamann: Okay! Yeah that was the thing I was deciding between and I wasn't sure how we wanted to handle it. I didn't know if we wanted to be able to configure how the cookie consent is handled. If we're sure we want google analytics on each site that uses the theme, then I'm definitely pro letting less configuration and having the component be very opinionated.

packages/gatsby-theme-newrelic/README.md
caylahamann: Actually I think it's neither of these, I think it is `its` 😆

packages/gatsby-theme-newrelic/src/components/CookieConsentDialog.js
caylahamann: I agree! I'll make those changes

packages/gatsby-theme-newrelic/src/components/CookieConsentDialog.js
caylahamann: I think it's good to make it very opinionated now, and then to change later in case we need to support different handling of the cookies.

packages/gatsby-theme-newrelic/gatsby-config.js
caylahamann: okay! the only thing i'm not sure about is if the `gaTrackingId` is the same for all the sites, but if not then i'll configure it up front.

demo/src/pages/index.js
caylahamann: Yeah for some reason vscode hasn't been formatting my files on save and I've tried a bunch of things to fix it. I've just been running prettier from the command line.

packages/gatsby-theme-newrelic/README.md
jerelmiller: Ha you're right! Apparently my brain was not working when I suggested this haha.

packages/gatsby-theme-newrelic/src/components/CookieConsentDialog.js
```diff
@@ -0,0 +1,149 @@
+import React, { useState } from 'react';
+import PropTypes from 'prop-types';
+import Cookies from 'js-cookie';
+import { css } from '@emotion/core';
+import ExternalLink from './ExternalLink';
+import Button from './Button';
+
+const cookieName = 'newrelic-gdpr-consent';
```
jerelmiller, line 8: [nit] `COOKIE_NAME`

packages/gatsby-theme-newrelic/src/components/CookieConsentDialog.js
```diff
@@ -0,0 +1,149 @@
...
+
+  const writeCookie = (answer) => {
+    const currentEnvironment =
+      process.env.ENV || process.env.NODE_ENV || 'development';
+    const options = { expires: 365 /* days */ };
+    if (currentEnvironment !== 'development') {
+      options.domain = 'newrelic.com';
+    }
+
+    Cookies.set(cookieName, String(!!answer), options);
```
jerelmiller, line 21: [nit] It looks like the `answer` passed here is always a boolean, so no need for the `!!`. You can just `String(answer)`

packages/gatsby-theme-newrelic/README.md
```diff
@@ -230,6 +231,10 @@ module.exports = {
 };
 ```
 
+#### `gdprTracking`
```
jerelmiller, line 234: A couple things here:

1) This configuration is not used in this PR, so having this listed in the README make confuse others. That being said, if we could only make the `gaTrackingId` configurable, then it would be great to list that here.
2) Can you make sure this gets added to the TOC above?

packages/gatsby-theme-newrelic/README.md
```diff
@@ -558,6 +563,26 @@ const Documentation = () => (
...
+A dialog box that pops up asking for cookie consent. This component renders at the bottom of the screen and provides options for accepting or denying cookies.
+
+```js
+import { CookieConsentDialog } from '@newrelic/gatsby-theme-newrelic';
+```
+
+**Example**
+
+```js
+const myLayout = () => (
```
jerelmiller, line 577: [nit] `Layout`

Figure it probably best to use React names to ensure we communicate that this likely goes in a layout _component_ (which are capitalized)

jerelmiller (commented):
@caylahamann Had a few more suggestions. Would you mind opening taking a look at these? I'd love to see some of these addressed before we release the theme again so that we can include this in the next version. Thanks!!

PR #146, opened 2020-10-31
"Right rail" component

### LINKED ISSUE #144: Create a common right rail component that can be used across sites
## Description
The docs site will have a "right rail" component that will show a few page-specific items.

The component will look similar to the right rail that appears on the developer site, so we want to abstract it to the theme. This component will continue to utilize its modular architecture to allow for customization of what goes in the right rail. This will also allow site-specific modules that can be used in the right rail.

We also plan to abstract a couple of the common modules that we know are used on all the sites. This includes the contributing module and the last modified module.

NOTE: This is followup of https://github.com/newrelic/docs-website/issues/191 that is specific to the theme work. The docs-site work will be done as a part of https://github.com/newrelic/docs-website/issues/191..

## Acceptance Criteria
* [x] Generalized right-rail component added to the Gatsby theme
* [x] GitHub buttons link to the correct locations
* [x] Ability to set contributing guide (default to `CONTRIBUTING.md`)

## Screenshot(s)
**Right Rail on Developer Site:**
Screen Shot 2020-10-29 at 4 02 30 PM

**Wireframes for Docs Site:**
Screen Shot 2020-10-29 at 4 03 24 PM

jerelmiller:
Closes #144 

## Description

Adds a `PageTools` component that can be leveraged to show content as a "right rail" for each Gatsby site. This also includes some building blocks as well as a `ContributingGuidelines` component that can be used to display information on how to edit the current page/file an issue/link to contributing guidelines.

### DIFF
--- a/packages/gatsby-theme-newrelic/README.md
+++ b/packages/gatsby-theme-newrelic/README.md
@@ -580,14 +580,14 @@ Used to display contributing information for the current page. This is meant to
 be used as a section inside of the [`PageTools`](#pagetools) component.
 
 To ensure this component leverages its full potential, please ensure the
-following [`siteMetadata`](#sitemetadata) fields are configured:
+following [`siteMetadata`](#site-metadata) fields are configured:
 
 - `branch`
 - `contributingUrl`
 - `repository`
 
-**NOTE:** If the `contributingUrl` is not configured, it will use the default value as
-specified in the [`siteMetadata`](#sitemetadata) section.
+**NOTE:** If the `contributingUrl` is not configured, it will use the default
+value as specified in the [`siteMetadata`](#site-metadata) section.
 
 ```js
 import { ContributingGuidelines } from '@newrelic/gatsby-theme-newrelic'`
--- a/packages/gatsby-theme-newrelic/README.md
+++ b/packages/gatsby-theme-newrelic/README.md
@@ -17,6 +17,7 @@ websites](https://opensource.newrelic.com).
     - [`robots`](#robots)
     - [`layout`](#layout)
     - [`prism`](#prism)
+    - [`gdprTracking`](#gdprtracking)
     - [`splitio`](#splitio)
       - [Environment-specific configuration](#environment-specific-configuration)
 - [Components](#components)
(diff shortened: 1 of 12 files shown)


packages/gatsby-theme-newrelic/README.md
```diff
@@ -943,6 +985,85 @@ import { NewRelicLogo } from '@newrelic/gatsby-theme-newrelic';
...
+```js
+import { PageTools } from '@newrelic/gatsby-theme-newrelic';
+```
+
+**Props**
+
+| Prop        | Type   | Required | Default | Description                                          |
+| ----------- | ------ | -------- | ------- | ---------------------------------------------------- |
+| `className` | string | no       |         | Additional `className` for the component.            |
+| `children`  | node   | no       |         | Content to be displayed in the `PageTools` component |
```
roadlittledawn, line 1005: Do we want to explicitly note that Section contents takes HTML and not markdown?

packages/gatsby-theme-newrelic/src/components/GlobalHeader.js
```diff
@@ -98,7 +98,7 @@ const GlobalHeader = ({ className }) => {
       >
         <div
           css={css`
-            height: 36px;
+            height: var(--global-header-height);
```
roadlittledawn, line 101: i imagine we can eventually use this variable to offset the scroll to anchor link?

roadlittledawn (approved):
Looks good! Just had one minor question re: documenting `<PageTools.Section />`

packages/gatsby-theme-newrelic/README.md
jerelmiller: Technically `children` could be anything so it really depends on where the `PageTools` component is rendered. If this happened to be used in an MDX file (we shouldn't in practice) then markdown would render just fine. Since we don't call this out for any other component, I wouldn't specify this as one way or the other.

packages/gatsby-theme-newrelic/src/components/GlobalHeader.js
jerelmiller: Yep!

PR #165, opened 2020-12-14
Updates to the Feedback Component

zstix:
## Description
Adds three additional props to the `<Feedback />` component to allow for more flexibility in how sites can use it:
* `showComment` (defaults to `true`)
* `onPositiveClick` (new event handler)
* `onNegativeClick` (new event handler)

## Related Issue(s)
* Relates to https://github.com/newrelic/docs-website/issues/300

## Screenshot(s)
Screen Shot 2020-12-14 at 1 58 32 PM

### DIFF
--- a/demo/src/pages/index.js
+++ b/demo/src/pages/index.js
@@ -283,6 +283,8 @@ const IndexPage = () => {
                 console.log('comment', sentiment, comment);
               }}
               showComment={false}
+              onPositiveClick={() => console.log('this is great!')}
+              onNegativeClick={() => console.log('this is terrible!')}
             />
           </div>
         </section>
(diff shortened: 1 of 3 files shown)


jerelmiller (commented):
I like the changes, but since we will be linking to a GitHub issue, I'm wondering if it would make sense to make these links instead of a click handler. It helps with a right-click open in new tab, or command click to open in a new tab. Thoughts?

zstix:
@jerelmiller Yeah I agree with that. Perhaps we should just handle this in the docs site as it's own component rather than do a ton of work to make the `<Feedback />` component work as a pair of links, or two buttons with a textarea. What do you think?

jerelmiller:
I like that idea!

zstix:
Closing this. We're going to implement the docs-specific feedback solution in that repo!

PR #166, opened 2020-12-15
Added a `<SimpleFeedback />` Component

zstix:
## Description
Adds the [`<SimpleFeedback />`](https://github.com/newrelic/docs-website/pull/412) component to the theme so that multiple sites can make use of it. Clicking one of the sentiment buttons will open up a new browser tab and take the user to the new issue screen. Depending on the props provided, different elements will be pre-filled and special labels are applied.

I also added some tests for this component, since there are a few different states the component can be in that are easily testable. I also wanted to provide us with an example of how to mock `useStaticQuery` and `graphql` from Gatsby.

## Screenshot(s)
Screen Shot 2020-12-15 at 10 00 00 AM

### DIFF
--- a/packages/gatsby-theme-newrelic/src/components/SimpleFeedback.js
+++ b/packages/gatsby-theme-newrelic/src/components/SimpleFeedback.js
@@ -64,6 +64,7 @@ const SimpleFeedback = ({ title, slug }) => {
           href={positiveFeedback}
           variant={Button.VARIANT.NORMAL}
           target="_blank"
+          role="button"
         >
           <Icon
             size="0.75rem"
@@ -79,6 +80,7 @@ const SimpleFeedback = ({ title, slug }) => {
           href={negativeFeedback}
           variant={Button.VARIANT.NORMAL}
           target="_blank"
+          role="button"
         >
           <Icon
             size="0.75rem"
--- a/packages/gatsby-theme-newrelic/src/components/SimpleFeedback.js
+++ b/packages/gatsby-theme-newrelic/src/components/SimpleFeedback.js
@@ -63,13 +63,16 @@ const SimpleFeedback = ({ title, slug, labels }) => {
           margin-top: 0.5rem;
           justify-content: center;
           align-items: flex-start;
+          gap: 0.5rem;
 
           a {
             flex-grow: 1;
           }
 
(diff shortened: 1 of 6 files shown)


packages/gatsby-theme-newrelic/src/components/SimpleFeedback.js
```diff
@@ -0,0 +1,112 @@
...
+  const issueTitle = title
+    ? `Feedback:+${title.replace(' ', '+')}`
+    : 'Website Feedback';
+
+  const body =
+    title && slug ? `&body=Page:%20[${title}](${siteUrl}${slug})` : '';
+
+  const positiveFeedback = `${issueUrl}?labels=${labels.join(
+    ','
+  )},feedback-positive&title=${issueTitle}${body}`;
```
jerelmiller, line 34: In the event that `labels` is empty, this might add an extra comma that you don't intend. You might consider appending the default label to the `labels` array and then `.join`

```js
`labels=${[...labels, 'feedback-negative'].join(',')}
```

packages/gatsby-theme-newrelic/src/components/SimpleFeedback.js
```diff
@@ -0,0 +1,112 @@
...
+          display: flex;
+          margin-top: 0.5rem;
+          justify-content: center;
+          align-items: flex-start;
+
+          a {
+            flex-grow: 1;
+          }
+
+          > *:first-child {
```
jerelmiller, line 60: [nit] You can totally leverage `gap` if you want to use it in browsers that support it! You just have to use the [`@supports` at-rule](https://developer.mozilla.org/en-US/docs/Web/CSS/@supports).

```js
css`
  gap: 0.5rem;

  @supports not (gap: 0.5rem) {
    > *:first-child {
      margin-right: 0.5rem;
    }
  }
`
```

jerelmiller (commented):
Added 2 minor comments. Let me know what you think!

packages/gatsby-theme-newrelic/src/components/SimpleFeedback.js
zstix: Good catch!

jerelmiller (approved):
🚢

PR #183, opened 2020-12-29
feat: added icon prop

jpvajda:
added icon prop to buttons

### DIFF
--- a/packages/gatsby-theme-newrelic/src/components/Button.js
+++ b/packages/gatsby-theme-newrelic/src/components/Button.js
@@ -2,6 +2,7 @@ import PropTypes from 'prop-types';
 import { css } from '@emotion/core';
 import styled from '@emotion/styled';
 import { rgba } from 'polished';
+import icon from './Icon.js';
 
 const VARIANTS = {
   PRIMARY: 'primary',
@@ -14,6 +15,10 @@ const VARIANTS = {
 const SIZES = {
   EXTRA_SMALL: 'extraSmall',
   SMALL: 'small',
+};
+
+const ICONS = {
+  // need something here....
 };
 
 const styles = {
@@ -134,11 +139,13 @@ const Button = styled.button`
 
 Button.VARIANT = VARIANTS;
 Button.SIZE = SIZES;
+Button.ICON = ICONS;
 
 Button.propTypes = {
   disabled: PropTypes.bool,
   size: PropTypes.oneOf(Object.values(Button.SIZE)),
   variant: PropTypes.oneOf(Object.values(Button.VARIANT)).isRequired,
+  icon: PropTypes.oneOf(Object.values(Button.ICON)),
 };
 
 export default Button;


packages/gatsby-theme-newrelic/src/components/Button.js
```diff
@@ -16,6 +17,10 @@ const SIZES = {
   SMALL: 'small',
 };
 
+const ICONS = {
+  // need something here....
```
caylahamann, line 21: i think you will want to add something like this: 
```js
const ICONS = {
LINK: 'external-link'
}
```

packages/gatsby-theme-newrelic/src/components/Button.js
```diff
@@ -134,11 +139,13 @@ const Button = styled.button`
```
caylahamann, line 139: so we would add a little more code here. We would need to render the button and then apply all the classes (right now it just returns the styled button, but since you are providing an icon you will need to render that within the button. So for example, you will need something like this 

```js 
const Button = ({variant, size, disabled, icon, children}) => (
<button css={css`/*all the styles here*/`}>
    { icon ? <Icon name={icon}/> : null } 
    { children }
</button>
)
```

caylahamann: this would replace the styled.div above

zstix (commented):
Rather than coupling our `<Button />` and `<Icon />` components, we should probably just compose them as needed:
```jsx
<Button>
  <Icon name="thumbs-up" />
  <span>Nice</span>
</Button>
```

jpvajda:
@zstix I assume you mean we want to be able to do something like this for a button 

```
 <Button
  role="button"
  as={Link}
  to="https://one.newrelic.com/launcher/nr1-core.settings?pane=eyJuZXJkbGV0SWQiOiJ0dWNzb24ucGxnLWluc3RydW1lbnQtZXZlcnl0aGluZyJ9&cards[0]=eyJuZXJkbGV0SWQiOiJzZXR1cC1uZXJkbGV0cy5zZXR1cC1qYXZhLWludGVncmF0aW9uIiwiYWNjb3VudElkIjoyNjQwNDA5fQ==&platform[accountId]=1"
  variant="primary"
  icon="external-link"
>
  Add Java data
</Button>
```

Where icon is set to the icon we wish to show. 

This page has the button in question: http://localhost:8000/docs/agents/java-agent/getting-started/introduction-new-relic-java

zstix:
@jpvajda No, I am suggesting that we _do not_ add the `icon` prop the the `<Button />` component. We have an `<Icon />` component and I think we should always strive for composition over "baking in" functionality.

Going this route, your example would be this (don't forget the shortened URLs and use the exported button / icon properties):
```jsx
<Button
  role="button"
  as={Link}
  to="https://one.nr/0VKQXA2kzwa"
>
  <Icon name={Icon.TYPE.EXTERNAL_LINK} />
  <span>Add Java data</span>
</Button>
```

jpvajda:
@zstix  aha! Good call, also great catch on the shorten link...

jpvajda:
It sounds like we don't need to make a theme change to support this, so I'll close this PR for now.

PR #184, opened 2020-12-30
feat: added signup button logic

jpvajda:
this PR adds logic to the sign up button to show sign up text when on the docs site domains, and start now text on all other domains.  To test you can use `window.location.href` as `window.location.hostname` doesn't work on localhost URLS.

```
//testing

const domain = window.location.href;

console.log(window.location.href);

{domain === 'http://localhost:8001/' ||
domain === 'http://localhost:8001/build-apps'
  ? 'Sign Up'
  : 'Start Now'}
```

### DIFF
--- a/packages/gatsby-theme-newrelic/src/components/GlobalHeader.js
+++ b/packages/gatsby-theme-newrelic/src/components/GlobalHeader.js
@@ -50,6 +50,7 @@ const GlobalHeader = ({ className }) => {
       site {
         siteMetadata {
           utmSource
+          siteUrl
         }
         layout {
           contentPadding
@@ -60,7 +61,7 @@ const GlobalHeader = ({ className }) => {
   `);
 
   const {
-    siteMetadata: { utmSource },
+    siteMetadata: { utmSource, siteUrl },
     layout,
   } = site;
 
@@ -77,7 +78,11 @@ const GlobalHeader = ({ className }) => {
 
   const hideLogoText = useMedia({ maxWidth: '655px' });
   const useSearchIcon = useMedia({ maxWidth: '585px' });
-  const domain = window.location.hostname;
+  // const domain = window.location.hostname;
+  const inDocsSite = [
+    'https://docs.newrelic.com',
+    'https://docs-preview.newrelic.com',
+  ].includes(siteUrl);
 
   return (
     <>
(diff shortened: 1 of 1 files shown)


packages/gatsby-theme-newrelic/src/components/GlobalHeader.js
```diff
@@ -77,6 +77,7 @@ const GlobalHeader = ({ className }) => {
 
   const hideLogoText = useMedia({ maxWidth: '655px' });
   const useSearchIcon = useMedia({ maxWidth: '585px' });
+  const domain = window.location.hostname;
```
zstix, line 80: We wont hav access to `window` at build time. When a site that uses this code builds, it will be in a purely node environment (not in a browser) so there's no concept of a window.

That said, we should be able to query for it using GraphQL. If you add `siteUrl` to the query above (in the `siteMetada` block), that should have the information you're looking for.

packages/gatsby-theme-newrelic/src/components/GlobalHeader.js
```diff
@@ -289,7 +290,12 @@ const GlobalHeader = ({ className }) => {
                 size={Button.SIZE.EXTRA_SMALL}
                 variant={Button.VARIANT.PRIMARY}
               >
-                <span>Sign up</span>
+                <span>
+                  {domain === 'www.docs-preview.newrelic.com' ||
+                  domain === 'www.docs.newrelic.com'
+                    ? 'Sign Up'
+                    : 'Start Now'}
```
zstix, line 297: An alternative using `includes` (and moving the logic out of the JSX) might make this a little more succinct:

```jsx
// above the JSX
const inDocsSite = ['https://developer.newrelic.com', 'https://docs-preview.newrelic.com'].includes(siteUrl);

// in the JSX
<span>{isDocSite ? 'Sign Up' : 'Start Now'}</span>
```

zstix (changes requested):
Mostly minor stuff, the nitpicky comment about `includes` can be ignored, but we will want to use the alternative to `window` so that our builds don't break.

packages/gatsby-theme-newrelic/src/components/GlobalHeader.js
```diff
@@ -77,6 +78,11 @@ const GlobalHeader = ({ className }) => {
 
   const hideLogoText = useMedia({ maxWidth: '655px' });
   const useSearchIcon = useMedia({ maxWidth: '585px' });
+  // const domain = window.location.hostname;
```
zstix, line 81:
```suggestion
```

zstix (approved):
🎉

PR #185, opened 2020-12-30
Adds `gatsby-source-swiftype` to theme

zstix:
## Description
This PR copies the `gatsby-source-swiftype` plugin [that we created for the developer site](https://github.com/newrelic/developer-website/pull/594) to the theme and exposes an optional `swiftype` option that can be passed to the theme by any site using the theme. This will allow us to use this logic for both the developer site and the new docs site.

## Reviewer Notes
I'm not 100% that we can have local plugins in a theme like this. Once this is merged and released, I'll work on swapping the developer site over to use the plugin via the theme (I'll link that PR here for transparency).

## Related Issue(s)
* Relates to https://github.com/newrelic/docs-website/issues/427


packages/gatsby-theme-newrelic/plugins/gatsby-source-swiftype/gatsby-node.js
```diff
@@ -0,0 +1,93 @@
...
+  resources.forEach((resource) => {
+    const child = createRelatedResourceNode({
+      parent: node.id,
+      resource,
+      createContentDigest,
+      createNode,
+      createNodeId,
+    });
+
+    createParentChildLink({ parent: node, child: child });
```
caylahamann, line 46: I'm guessing that in order to get the related resources we'll query for the children of the node?

packages/gatsby-theme-newrelic/plugins/gatsby-source-swiftype/gatsby-node.js
```diff
@@ -0,0 +1,93 @@
...
+    fs.writeFileSync(file, '{}');
+  }
+};
+
+exports.onCreateNode = async (
+  { actions, node, getNodesByType, createNodeId, createContentDigest },
+  pluginOptions
+) => {
+  const { createNode, createParentChildLink } = actions;
+  const { filterNode = () => false, getPath } = pluginOptions;
```
caylahamann, line 20: what does this `filterNode` function do? how does this inline function work

caylahamann (changes requested):
this all looks good to me! just two clarifying questions, and the one suggestion that I have is that we change the name of the plugin to `gatsby-source-related-resources`? It just makes sense to me since we are sourcing the related resources, but feel free to challenge me on this!

packages/gatsby-theme-newrelic/plugins/gatsby-source-swiftype/gatsby-node.js
zstix: The PR is just copying what we had before. I didn't write the original code, but I believe we need to establish a relationship so that we can query related content for a specific page.

packages/gatsby-theme-newrelic/plugins/gatsby-source-swiftype/gatsby-node.js
zstix: It looks like this is being used as a default value in case `filterNode` isn't specified. If we use this plugin without `filterNode`, it will just be a function that always returns `false`.

zstix:
@caylahamann Let me know if you have any additional questions on this PR.

cc: @jerelmiller (this PR brings the work you did on the developer site into the theme so that we have the _option_ of getting related content on other sites)

packages/gatsby-theme-newrelic/plugins/gatsby-source-swiftype/gatsby-node.js
jerelmiller: Yes. This creates a `childrenRelatedPages` (if I remember correctly) field on the parent so that you can query for it for the current page.

jerelmiller (approved):

jerelmiller:
TODO: Write docs

packages/gatsby-theme-newrelic/plugins/gatsby-source-swiftype/gatsby-node.js
jerelmiller: The reason that `filterNode` returns `false` rather than `true` is that we want to avoid executing searches on Swiftype for pages that aren't explicitly set as needing related resources. This helps us avoid the cost of executing a query against Swiftype for pages that we don't use this for.

caylahamann (approved):

PR #191, opened 2021-01-06
Adds the japanese pages based on file relative path (in src/pages directory) 

caylahamann:


jerelmiller (approved):
I think this is a good solution. Maybe we can make it more robust later, but this works great for now! Thanks for coming back to this.

PR #194, opened 2021-01-07
Setup i18next framework for the theme

jerelmiller:
## Description

Sets up `react-i18next` for use from the theme. This will allow us to not only translate the existing components in the theme, but include translations from the sites themselves. Re-exports all modules from `react-i18next` so that the consumer does not have to install the library themself.


packages/gatsby-theme-newrelic/gatsby-node.js
```diff
@@ -140,26 +170,80 @@ exports.onCreateNode = ({ node, actions }) => {
...
+  if (fs.existsSync(filepath)) {
+    return;
+  }
+
+  if (message) {
+    reporter.info(message);
+  }
+
+  createDirectory(path.dirname(filepath));
+  fs.writeFileSync(filepath, '{}', 'utf-8');
```
jerelmiller, line 245: make the content an argument

zstix (approved):
This looks awesome.

That was one heck of a `reduce`

jerelmiller:
Thank you!!

PR #198, opened 2021-01-09
Use localized text for all components + localized link

### LINKED ISSUE #180: Transform gatsby components to render react-intl context
## Summary 
We need to set up `gatsby-plugin-intl` on our gatsby theme or provide a way to render context for the global header and footer, as well as a few other components used by our docs site. 

## AC 
* [x] Gatsby theme allows for language context switching on our sites 
* [x] Global header internationalized
* [ ] Global footer internationalized 
* [ ] Code block internationalized
* [ ] Feedback internationalized
* [ ] Cookie Consent Dialog internationalized

jerelmiller:
Closes #180 

## Description

Uses localized text for all components. Adds a utility that allows us to easily extract all keys into the translation file. This PR also updates the `Link` component to work as a localized link.


demo/src/layouts/index.js
```diff
@@ -1,6 +1,21 @@
...
+const Layout = (props) => {
+  const { children, pageContext } = props;
+
+  switch (pageContext.layout) {
+    case 'basic':
+      return children;
+    default:
+      return <MainLayout {...props} />;
+  }
+};
```
zstix, line 14: Question: do we suspect there will be other cases where we would like to render layout differently? Curious why you reached for a `switch` here vs an early return.

demo/src/layouts/index.js
```diff
@@ -1,6 +1,21 @@
...
+
+Layout.propTypes = {
+  children: PropTypes.node,
+  pageContext: PropTypes.object.isRequired,
```
zstix, line 18: Do we want to add a shape for this that includes `layout`?

packages/gatsby-theme-newrelic/gatsby/wrap-page-element.js
```diff
@@ -22,12 +23,34 @@ const wrapPageElement = ({ element, props }, themeOptions) => {
...
+
+  // For some reason we enter an infinite loop if we try and create localized
+  // 404 pages in gatsby-node.js. This is a hack that will use the current page
+  // location to find the locale to ensure the 404 page is translated
+  // accordingly.
+  if (!pageContext.fileRelativePath.match(/404/)) {
+    return pageContext.locale || defaultLocale.locale;
+  }
+
+  const [, base] = location.pathname.split('/');
```
zstix, line 46: Woah, didn't know about `[, thing]` being valid...I've been doing `[_, thing]` like a fool!

zstix (commented):
Couple of small questions, but this is looking 💯

demo/src/layouts/index.js
jerelmiller: This is the demo site. To be honest, I didn't put much thought into it. I just copied it from here:

https://github.com/newrelic/docs-website/blob/3e9e090016d4084546c49365d0763591c0aee468/src/layouts/index.js#L8-L13

demo/src/layouts/index.js
jerelmiller: Since this is the demo site, I wasn't too careful on proper prop types 😄 . I'd be happy to add it if you feel strongly about it.

packages/gatsby-theme-newrelic/gatsby/wrap-page-element.js
jerelmiller: Haha its a great trick! The nice thing about `[,]` is that you don't need to waste a variable name!

demo/src/layouts/index.js
jerelmiller: Ultimately this will be implemented by the sites themselves. This layout is only for the demo site, which has its own layout (which currently mimics the current docs/dev sites)

demo/src/layouts/index.js
zstix: I feel like I go on autopilot and forget to look at the filepath. Totally fine in the demo 👍

zstix (approved):

demo/src/layouts/index.js
jerelmiller: No problem!

PR #203, opened 2021-01-11
Update `pluginOptions` -> `themeOptions`

zstix:
## Description
When I brought the plugin for sourcing swiftype into the theme itself, I forgot to update all references of `pluginOptions` -> `themeOptions` (#201). This PR updates that and ensures we're using a consistent variable name throughout `gatstby-node.js`.

### DIFF
--- a/packages/gatsby-theme-newrelic/gatsby-node.js
+++ b/packages/gatsby-theme-newrelic/gatsby-node.js
@@ -61,7 +61,7 @@ exports.onPreBootstrap = ({ reporter, store }, themeOptions) => {
   }
 
   if (themeOptions.swiftype) {
-    const { file } = pluginOptions.swiftype;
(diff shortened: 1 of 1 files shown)


jerelmiller (approved):
Would you mind changing your commit to a `fix` so that it shows up in the changelog?

zstix:
@jerelmiller good call, I've updated that commit. Always makes me nervous updating commits when they're already pushed up, but this feels fairly safe.

jerelmiller:
As long as you're using `--force-with-lease`, you can make sure that you have all the changes synced from remote to your local before a force push has the potential to overwrite code. I wouldn't stress about it too much unless someone else is actively working on the same branch. Anyways, thanks for updating the commit!

jerelmiller (approved):

PR #207, opened 2021-01-11
Fixes for fetching related content

zstix:
## Description
Makes a few more adjustments to update the swiftype code to work in the theme. Includes supplying a fallback for the `getPath` option.

### DIFF
--- a/packages/gatsby-theme-newrelic/gatsby-node.js
+++ b/packages/gatsby-theme-newrelic/gatsby-node.js
@@ -1,5 +1,6 @@
 const fs = require('fs');
 const path = require('path');
+const { createFilePath } = require('gatsby-source-filesystem');
 const {
   defaultLocale,
   withDefaults,
@@ -215,7 +216,10 @@ exports.onCreateBabelConfig = ({ actions }, themeOptions) => {
   });
 };
 
-exports.onCreateNode = async ({ node, actions }, themeOptions) => {
+exports.onCreateNode = async (
+  { node, actions, getNode, getNodesByType, createContentDigest, createNodeId },
+  themeOptions
+) => {
   const {
     createNode,
     createNodeField,
@@ -231,7 +235,7 @@ exports.onCreateNode = async ({ node, actions }, themeOptions) => {
     });
   }
 
-  if (themeOptions.swiftype && getNodesByType) {
+  if (themeOptions.swiftype) {
     const { filterNode = () => false, getPath } = themeOptions.swiftype;
 
     if (node.internal.type !== 'Mdx' || !filterNode({ node })) {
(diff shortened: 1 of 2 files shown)


packages/gatsby-theme-newrelic/gatsby-node.js
```diff
@@ -320,11 +326,9 @@ exports.onCreateWebpackConfig = ({ actions, plugins }, themeOptions) => {
 };
 
 exports.onPostBootstrap = (_, themeOptions) => {
-  const { refetch, file } = themeOptions;
-
-  if (refetch) {
+  if (themeOptions.swiftype && themeOptions.swiftype.refetch) {
     fs.writeFileSync(
-      file,
+      themeOptions.swiftype.file,
```
jerelmiller, line 331: Now that I see this in the theme, should we add a validation in `onPreInit` to make sure this option is given if `swiftype` is defined? We'll need to make sure this file exists.

We may also want something in `onPreBootstrap` to create this file for you if its not already been created.

packages/gatsby-theme-newrelic/gatsby-node.js
```diff
@@ -320,11 +326,9 @@ exports.onCreateWebpackConfig = ({ actions, plugins }, themeOptions) => {
 };
 
 exports.onPostBootstrap = (_, themeOptions) => {
-  const { refetch, file } = themeOptions;
-
-  if (refetch) {
+  if (themeOptions.swiftype && themeOptions.swiftype.refetch) {
```
jerelmiller, line 329: If you hate typing `themeOptions.swiftype` over and over, you should be able to use a combo of destructuring and defaults to make this a bit easier to parse. Up to you if you want to do this:

```js
const { swiftype = {} } = themeOptions

if (swiftype.refetch) {
  fs.writeFileSync(swiftype.file, ....)
}
```

packages/gatsby-theme-newrelic/src/utils/related-resources/fetchRelatedResources.js
```diff
@@ -1,18 +1,15 @@
 const fs = require('fs');
 const search = require('./search');
 
-module.exports = async ({ node, siteUrl }, pluginOptions) => {
+module.exports = async ({ pathname, node, siteUrl }, pluginOptions) => {
```
jerelmiller, line 4:
```suggestion
module.exports = async ({ pathname, node, siteUrl }, swiftypeOptions) => {
```

Could we rename this so that we don't confuse this with the theme options? This makes it a bit more clear that these aren't just the full set of options passed to the theme.

packages/gatsby-theme-newrelic/gatsby-node.js
zstix: I'll just add another conditional check for now, since I we need to get the fixes in this PR released soon.

caylahamann (approved):
reviewing this and looking at jerel's suggestions, it all looks good to me!

PR #208, opened 2021-01-12
SEO component in theme with hreflang tags added 

caylahamann:
## Description
Pulls the SEO component out into the theme, and additionally renders hreflang tags into the component

Closes newrelic/docs-website#502


packages/gatsby-theme-newrelic/src/components/SEO.js
```diff
@@ -0,0 +1,62 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import Helmet from 'react-helmet';
+import { graphql, useStaticQuery } from 'gatsby';
+import useLocale from '../hooks/useLocale';
+const path = require('path');
+
+const SEO = ({ title, location }) => {
```
jerelmiller, line 8: We should also add a `children` prop to allow for individual sites to add additional values. For example, the developer site adds a bunch of Swiftype related stuff here: https://github.com/newrelic/developer-website/blob/develop/src/components/Seo.js#L45-L76

caylahamann:
FYI: going to add documentation after I review if there's anything else needed from the developer site or open source site in terms of SEO functionality

packages/gatsby-theme-newrelic/src/components/SEO.js
```diff
@@ -0,0 +1,62 @@
...
+      }
+    }
+  `);
+
+  const currentLocale = useLocale();
+
+  const { defaultTitle } = siteMetadata;
+
+  const subPath =
+    currentLocale === 'en'
```
jerelmiller, line 33: Rather than hardcoding `en` here, could you instead query for `isDefault` and find the default locale? While we are using `en` right now, if this value ever changes, its going to be tough to grep for. By finding the default locale, we can ensure that we are always using the configured value, which makes changing the default easy. 

```js
useStaticQuery(graphql`
  allLocale {
    nodes {
      locale
      localizedPath
      isDefault
  }
`)

const defaultLocale = locales.find(({ isDefault }) => isDefault)

currentLocale.locale === defaultLocale.locale?
```

jerelmiller: On that note, the `curentLocale` you pull in here is a full `Locale` node. You'll need to compare `currentLocale.locale` to ensure you're comparing string values.

jerelmiller (changes requested):

packages/gatsby-theme-newrelic/src/components/SEO.js
caylahamann: I was thinking along those lines as well! I like that idea, makes it easier to migrate over / if we ever want to abstract anything back into the theme

packages/gatsby-theme-newrelic/src/components/SEO.js
caylahamann: yes! will do that, and thank you for pointing that out. I will make that change

jerelmiller (approved):
Will re-approve after you've been able to add documentation. Thanks for making those changes!

packages/gatsby-theme-newrelic/README.md
```diff
@@ -1583,6 +1584,36 @@ const Search = () => (
...
+
+```js
+import { SEO } from '@newrelic/gatsby-theme-newrelic';
+```
+
+**Props**
+
+| Prop       | Type   | Required | Default | Description                                                   |
+| ---------- | ------ | -------- | ------- | ------------------------------------------------------------- |
+| `location` | string | yes      |         | object which must have properties of `pathname` and `origin`. |
```
jerelmiller, line 1599:
```suggestion
| `location` | string | yes      |         | The location object of the page where this is rendered. This object must have properties of `pathname` and `origin`. |
```

packages/gatsby-theme-newrelic/README.md
```diff
@@ -1583,6 +1584,36 @@ const Search = () => (
...
+| `title`    | string | no       |         | title to be displayed in head.                                |
+| `children` | node   | no       |         | any additional tags to be added to the `<head>` of the page   |
+
+**Example**
+
+```js
+const MainLayout = () => (
+  return (
+    <>
+      <SEO location={location}/>
```
jerelmiller, line 1609:
```suggestion
      <SEO location={location} />
```

jerelmiller (commented):
2 minor suggestions on the documentation, but otherwise looks great!

jerelmiller (approved):

PR #218, opened 2021-01-15
Global style updates

jerelmiller:
## Description

* Refactor global styles to split up the CSS variables into their own folder. 
* Add CSS variables for theme layout config. 
* Add some additional used global styles for all sites. 
* Remove secondary and tertiary fonts


zstix (approved):
Thanks for cleaning this stuff up - looks great 🎨

PR #225, opened 2021-01-22
Document Collapser/CollapserGroup + other changes

jerelmiller:
[Rendered](https://github.com/newrelic/gatsby-theme-newrelic/blob/jerel/document-new-stuff/packages/gatsby-theme-newrelic/README.md)

### DIFF
--- a/packages/gatsby-theme-newrelic/README.md
+++ b/packages/gatsby-theme-newrelic/README.md
@@ -2008,7 +2008,18 @@ import { useKeyPress } from '@newrelic/gatsby-theme-newrelic';
 
 **Arguments**
 
-- `key` _(string)_: The key being listened for (i.e. the `event.key` value)
+- `key` _(string | Array<string>)_: The key or keys being listened for
+  (i.e. the `event.key` value). This hook also recognizes modifier keys. To add
+  a listener for a modifier key, use the format `<modifier key>+<key>` (i.e.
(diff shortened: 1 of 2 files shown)


packages/gatsby-theme-newrelic/README.md
```diff
@@ -656,6 +658,80 @@ const Documentation = () => (
...
+
+```js
+import { Collapser } from '@newrelic/gatsby-theme-newrelic'`
+```
+
+**Props**
+
+| Prop          | Type                    | Required | Default | Description                                                                                                                  |
+| ------------- | ----------------------- | -------- | ------- | ---------------------------------------------------------------------------------------------------------------------------- |
+| `children`    | enum                    | yes      |         | The content that will be hidden or revealed when the user interacts with the `Collapser`.                                    |
```
zstix, line 675: Shouldn't `children` be of the `node` type? If not, we should update the propTypes in the component.

zstix (commented):
Looks great, just one minor question!

packages/gatsby-theme-newrelic/README.md
jerelmiller: Oops, good catch. I copy/pasted this from another component and forgot to update this one.

jerelmiller: Fixed in 94a8519

zstix (approved):

PR #232, opened 2021-01-27
MDX component + MDX variants

jerelmiller:
Closes https://github.com/newrelic/docs-website/issues/262

# Description

Create an `MDX` component that will be used to render MDX content. This component is preloaded with shared components used in MDX documents.

This PR also adds MDX variants to all block-level components to provide proper spacing around them in an MDX document where spacing is difficult to control.


package.json
```diff
@@ -6,6 +6,7 @@
     "packages/*"
   ],
   "scripts": {
+    "start": "yarn workspace demo develop",
```
zstix, line 9: 💯

packages/gatsby-theme-newrelic/src/components/MDX.js
```diff
@@ -0,0 +1,46 @@
...
+  return (
+    <MDXProvider components={{ ...defaultComponents, ...components }}>
+      <MDXRenderer>{body}</MDXRenderer>
+    </MDXProvider>
+  );
+};
+
+MDX.propTypes = {
+  body: PropTypes.string.isRequired,
+  components: PropTypes.object,
```
zstix, line 43: Out of curiosity, is `components` open-ended, or do we only support specific components?

packages/gatsby-theme-newrelic/src/components/MarkdownContainer.js
```diff
@@ -0,0 +1,45 @@
...
+
+const MarkdownContainer = ({
+  children,
+  className,
+  dangerouslySetInnerHTML,
+}) => {
+  return (
+    <div
+      dangerouslySetInnerHTML={dangerouslySetInnerHTML}
+      className={className}
```
zstix, line 13: Since the keys equal the tag attributes, we could consider writing it like this:

```jsx
const MarkdownContainer = ({ children, ...props }) => (
  <div
    ...props
    css={`...`}
  >
    {children}
  </div>
);
```

That said, what you have hear is more descriptive. Just a thought.

zstix (approved):
I _really_ like this stuff! Gonna make working with MDX much more streamlined.

I had a couple comments / questions, but nothing I would block the PR for.

packages/gatsby-theme-newrelic/src/components/MDX.js
jerelmiller: `components` are open ended. This is just to provide additional shortcodes needed for MDX documents in the specific sites. For example, we have a `LandingPageTile` shortcode on the docs site, which the theme does not know about. This prop gives us the ability to map that on the docs end and merge with these default components.

packages/gatsby-theme-newrelic/src/components/MarkdownContainer.js
jerelmiller: Thats true. Really the only difference between them is one style allows _all_ props (`onClick`, etc) vs this just allowing these 2 components.

I don't feel strongly either way. I'd say lets keep this for now and when the need arises to add support for more props, we can switch over to the spread props. 

How do you feel about that?

PR #238, opened 2021-02-01
chore: sign up button > free account

jpvajda:
This PR removes the logic to handle the different buttons on each site, and makes all sign up buttons say Free Account.

### DIFF
--- a/packages/gatsby-theme-newrelic/src/components/GlobalHeader.js
+++ b/packages/gatsby-theme-newrelic/src/components/GlobalHeader.js
@@ -322,7 +322,7 @@ const GlobalHeader = ({ className }) => {
                 size={Button.SIZE.EXTRA_SMALL}
                 variant={Button.VARIANT.PRIMARY}
               >
-                <span> {t('button.signUp')}</span>
+                <span>{t('button.signUp')}</span>
               </Button>
             </li>
           </ul>
--- a/packages/gatsby-theme-newrelic/src/components/GlobalHeader.js
+++ b/packages/gatsby-theme-newrelic/src/components/GlobalHeader.js
@@ -75,10 +75,6 @@ const GlobalHeader = ({ className }) => {
 
   const hideLogoText = useMedia({ maxWidth: '655px' });
   const useCondensedHeader = useMedia({ maxWidth: '585px' });
-  const inDocsSite = [
-    'https://docs.newrelic.com',
-    'https://docs-preview.newrelic.com',
-  ].includes(siteUrl);
 
   const matchLocalePath = new RegExp(
     `^\\/(${locales.map(({ locale }) => locale).join('|')})`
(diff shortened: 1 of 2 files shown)


packages/gatsby-theme-newrelic/src/components/GlobalHeader.js
```diff
@@ -75,10 +75,6 @@ const GlobalHeader = ({ className }) => {
 
   const hideLogoText = useMedia({ maxWidth: '655px' });
   const useCondensedHeader = useMedia({ maxWidth: '585px' });
-  const inDocsSite = [
-    'https://docs.newrelic.com',
-    'https://docs-preview.newrelic.com',
-  ].includes(siteUrl);
```
zstix, line 81: If we're no longer relying on `siteUrl`, then we can probably remove it from the query above (line 59 and 73).

zstix (commented):
Looks great, just one minor comment!

packages/gatsby-theme-newrelic/src/components/GlobalHeader.js
```diff
@@ -327,9 +323,7 @@ const GlobalHeader = ({ className }) => {
                 size={Button.SIZE.EXTRA_SMALL}
                 variant={Button.VARIANT.PRIMARY}
               >
-                <span>
-                  {inDocsSite ? t('button.signUp') : t('button.startNow')}
-                </span>
+                <span>{'Free Account'}</span>
```
jerelmiller, line 326: We should make sure this ends up in our translation file. Rather than using the raw text `Free Account` here (which also should be sentence case), case you instead use `button.signUp` and change the translation in `src/i18n/translations/en.json`?

jerelmiller: Oh, and we should also be able to delete the `button.startNow` value in `en.json`

jpvajda: will do

packages/gatsby-theme-newrelic/src/components/GlobalHeader.js
jpvajda: will do

packages/gatsby-theme-newrelic/src/components/GlobalHeader.js
```diff
@@ -327,9 +322,7 @@ const GlobalHeader = ({ className }) => {
                 size={Button.SIZE.EXTRA_SMALL}
                 variant={Button.VARIANT.PRIMARY}
               >
-                <span>
-                  {inDocsSite ? t('button.signUp') : t('button.startNow')}
-                </span>
+                <span> {t('button.signUp')}</span>
```
jerelmiller, line 325:
```suggestion
                <span>{t('button.signUp')}</span>
```

jerelmiller (approved):

PR #260, opened 2021-02-06
chore: updated contributor CTA message

jpvajda:
This PR changes the contributor CTA in the right rail. 

Closes: https://github.com/newrelic/docs-website/issues/753

### DIFF
--- a/packages/gatsby-theme-newrelic/src/components/ContributingGuidelines.js
+++ b/packages/gatsby-theme-newrelic/src/components/ContributingGuidelines.js
@@ -101,8 +101,8 @@ const ContributingGuidelines = ({ fileRelativePath, pageTitle }) => {
             text-align: center;
           `}
         >
-          Read our <ExternalLink href={contributingUrl}>guide</ExternalLink> on
-          how to contribute
+          Suggest a change or learn how to
+          <ExternalLink href={contributingUrl}>contribute</ExternalLink>
         </Trans>
       )}
     </PageTools.Section>
--- a/packages/gatsby-theme-newrelic/src/components/ContributingGuidelines.js
+++ b/packages/gatsby-theme-newrelic/src/components/ContributingGuidelines.js
@@ -3,6 +3,7 @@ import PropTypes from 'prop-types';
 import { useLocation } from '@reach/router';
 import { css } from '@emotion/core';
 import Button from './Button';
+import Link from './Link';
 import ExternalLink from './ExternalLink';
 import Icon from './Icon';
 import PageTools from './PageTools';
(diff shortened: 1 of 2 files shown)


jerelmiller:
> as a note, I couldn't get this to render on local

I wonder if you're accessing the demo site on the wrong port. The demo site runs on port 8001 (not 8000) so that you can simultaneously develop on the theme while also having a "real" site running. Just an FYI :)

packages/gatsby-theme-newrelic/src/components/ContributingGuidelines.js
```diff
@@ -101,8 +101,8 @@ const ContributingGuidelines = ({ fileRelativePath, pageTitle }) => {
             text-align: center;
           `}
         >
-          Read our <ExternalLink href={contributingUrl}>guide</ExternalLink> on
-          how to contribute
+          Suggest a change or learn how to
+          <ExternalLink href={contributingUrl}>contribute</ExternalLink>
```
jerelmiller, line 105: Ah looks like one place where we are currently using `ExternalLink`. I'd like to deprecate that component if we can in favor or `Link` from the theme. `Link` has all of the same functionality built, so we can maintain a single component. Could you update this to use `Link`?

Bonus: By using `Link`, this will handle both if the `contributingUrl` is either external or internal. If its an internal link, then it will ensure it transitions nicely to the url in the site rather than opening the page in a new link.

NOTE: You'll need to [update the import](https://github.com/newrelic/gatsby-theme-newrelic/pull/260/files#diff-fd108bf279d260e105ada43a84784852755997a1d2c377531ad94c3a779ab780R6) to pull from `'./Link'` instead of `'./ExternalLink'`

```suggestion
          <Link to={contributingUrl}>contribute</Link>
```

jerelmiller (commented):
Had a minor suggestion for replacing `ExternalLink` with `Link`

jerelmiller:
One more minor thing: can you remove the [`ExternalLink` import](https://github.com/newrelic/gatsby-theme-newrelic/pull/260/files#diff-fd108bf279d260e105ada43a84784852755997a1d2c377531ad94c3a779ab780R7) now that you're using `Link`? It looks like that component is no longer used. ESLint might complain otherwise (speaking of which, it looks like we are missing that on the PR check 😅 )

jpvajda:
@jerelmiller external link is being used in the file though.. so I thought it was required.

jerelmiller:
D'oh! Totally missed that `as` prop on the `Button`. Thanks!

jerelmiller (approved):

PR #262, opened 2021-02-10
Add support for opening all collapsers via query param

roadlittledawn:
Closes [docs-website/795](https://github.com/newrelic/docs-website/issues/795)

## Description
Adds support for opening all collapsers via query param: `collapsers=openAll`. This is particularly useful when uploading contexts to Smartling so all content is visible to translators.

### DIFF
--- a/packages/gatsby-theme-newrelic/src/components/Collapser.js
+++ b/packages/gatsby-theme-newrelic/src/components/Collapser.js
@@ -1,10 +1,11 @@
-import React, { useState, useMemo } from 'react';
+import React, { useState, useMemo, useEffect } from 'react';
 import PropTypes from 'prop-types';
 import { css } from '@emotion/core';
 import Icon from './Icon';
 import { animated, useSpring } from 'react-spring';
 import { usePrevious, useIsomorphicLayoutEffect } from 'react-use';
 import useKeyPress from '../hooks/useKeyPress';
+import useQueryParams from '../hooks/useQueryParams';
 
 const ResizeObserver = global.ResizeObserver || class ResizeObserver {};
 
@@ -16,6 +17,15 @@ const Collapser = ({ title, id, defaultOpen, children }) => {
   const previousIsOpen = usePrevious(isOpen);
 
   useKeyPress(['s', 'f', 'h'], (e) => setIsOpen(e.key !== 'h'));
+
+  const { queryParams } = useQueryParams();
+
+  useEffect(() => {
+    setIsOpen(
+      queryParams.has('collapsers') &&
+        queryParams.get('collapsers') === 'showAll'
+    );
+  });
 
   const observer = useMemo(
     () =>
--- a/packages/gatsby-theme-newrelic/src/components/Collapser.js
+++ b/packages/gatsby-theme-newrelic/src/components/Collapser.js
@@ -10,22 +10,19 @@ import useQueryParams from '../hooks/useQueryParams';
 const ResizeObserver = global.ResizeObserver || class ResizeObserver {};
 
 const Collapser = ({ title, id, defaultOpen, children }) => {
+  const { queryParams } = useQueryParams();
   const [element, ref] = useState();
-  const [isOpen, setIsOpen] = useState(defaultOpen);
+  const [isOpen, setIsOpen] = useState(() => {
+    return queryParams.has('collapsers') &&
+      queryParams.get('collapsers') === 'openAll'
+      ? true
+      : defaultOpen;
+  });
   const [height, setHeight] = useState(0);
   const { height: viewHeight } = useSpring({ height: isOpen ? height : 0 });
   const previousIsOpen = usePrevious(isOpen);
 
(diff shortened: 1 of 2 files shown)


packages/gatsby-theme-newrelic/src/components/Collapser.js
```diff
@@ -1,16 +1,23 @@
-import React, { useState, useMemo } from 'react';
+import React, { useState, useMemo, useEffect } from 'react';
```
caylahamann, line 1:
```suggestion
import React, { useState, useMemo } from 'react';
```

caylahamann (approved):

packages/gatsby-theme-newrelic/src/components/Collapser.js
caylahamann: There also seems to be some tests failing 🤔

packages/gatsby-theme-newrelic/src/components/Collapser.js
```diff
@@ -1,16 +1,23 @@
...
 
 const ResizeObserver = global.ResizeObserver || class ResizeObserver {};
 
 const Collapser = ({ title, id, defaultOpen, children }) => {
+  const { queryParams } = useQueryParams();
   const [element, ref] = useState();
-  const [isOpen, setIsOpen] = useState(defaultOpen);
+  const [isOpen, setIsOpen] = useState(() => {
+    return queryParams.has('collapsers') &&
+      queryParams.get('collapsers') === 'openAll'
```
jerelmiller, line 17:
```suggestion
      queryParams.get('collapsers') === 'open'
```

I'd suggest use using `open`, or maybe even `1` to signify "truthy" (which I've seen, for example GitHub allows "w=1" to ignore whitespace). Using a query param is usually signal that it applies to everything (hence why I'd at least drop "all" from the value)

packages/gatsby-theme-newrelic/src/components/Collapser.js
```diff
@@ -1,16 +1,23 @@
...
+      ? true
+      : defaultOpen;
```
jerelmiller, line 19: Out of curiosity, what would you consider to be the correct behavior in this case?

```js
<Collapser defaultOpen={false} />

// url collapsers=openAll
```

Should the collapser be open or closed in this case? I ask because I'm not sure whether the query param should be a fallback or the override. Thoughts?

jerelmiller (commented):
Couple small questions. I'd also be curious to see if this works on a build of the site.

packages/gatsby-theme-newrelic/src/components/Collapser.js
roadlittledawn: ah yeah i see that. looks like the jest test for Collapsers is erroring due to using the hook `useQueryParams` which uses `useLocation`. so maybe it sees that the Collapser has some code in it that looks at / relies on location, but test doesn't handle that? any idea how to address that in a jest test?

packages/gatsby-theme-newrelic/src/components/Collapser.js
roadlittledawn: I was thinking the query param is an override to what is otherwise the default (do not open collapsers). i think the most common case is we'll want to collapse them by default unless we have a special need to open them (like supporting smartling contexts)

jerelmiller:
@roadlittledawn this should help to fix the tests (Gatsby uses Reach Router under the hood): https://reach.tech/router/api/LocationProvider

The key is to use "memory history" as part of the router and wrap the component in the test with that `LocationProvider`. The memory history has the same API has browser history, but as it states, "its mostly used for tests". Try that out and see if you can get the tests to pass!

jerelmiller:
On that note, it might be good to write a test to verify that the collapser is open when the query param is present :)

packages/gatsby-theme-newrelic/src/components/Collapser.js
roadlittledawn: i added `<LocationProvider>` to the collapser test components and it doesn't fail now. not exactly sure if this is the optimal solution though?

packages/gatsby-theme-newrelic/src/components/Collapser.js
roadlittledawn: > @roadlittledawn this should help to fix the tests (Gatsby uses Reach Router under the hood): https://reach.tech/router/api/LocationProvider
> 
> The key is to use "memory history" as part of the router and wrap the component in the test with that `LocationProvider`. The memory history has the same API has browser history, but as it states, "its mostly used for tests". Try that out and see if you can get the tests to pass!

oh just saw this right after i pushed it up. will look and see if i missed anything but the tests succeed now. and yep, was thinking i'll add a test for query param as well.

roadlittledawn: gotcha. i went with `open`

jerelmiller (approved):

PR #265, opened 2021-02-12
Allow nav items to be filtered

jerelmiller:
Relates to https://github.com/newrelic/docs-website/issues/512

# Description

Add the ability to filter nav items by a search term. This PR also introduces a `TextHighlight` component to highlight matching search terms in a string.

## Screenshots

Screen Shot 2021-02-12 at 11 55 31 AM
Screen Shot 2021-02-12 at 12 06 35 PM


packages/gatsby-theme-newrelic/src/components/Navigation.js
```diff
@@ -0,0 +1,30 @@
...
+    <NavigationContext.Provider value={value}>
+      <nav role="navigation" aria-label="Navigation" className={className}>
+        {children}
+      </nav>
+    </NavigationContext.Provider>
+  );
+};
+
+Navigation.propTypes = {
+  children: PropTypes.node,
```
zstix, line 25: Since we have it listed as required in the docs, we should probably update it here (or just update the docs).

```suggestion
  children: PropTypes.node.isRequired,
```

zstix (approved):
🎉 Looks good, one minor nitpick, but nothing I would block a merge for

packages/gatsby-theme-newrelic/src/components/Navigation.js
jerelmiller: Oops, meant to check that and didn't. Nice catch!

zstix (approved):

Issue #254, opened 2021-02-04
[Google] Add google analytics custom events 


caylahamann:
## Summary 
To reach parity with the current docs site, we need to send custom google events with GA. We need to track two events (that will be also instrumented with NR browser) `copyCodeBlock_click` and `externalLink_click`. They will send the same info that is sent with NR browser. We can implement this in one of two ways. 

### Option 1: Use global `ga` methods 
We can still access to the global `ga` methods to send custom events, we just won't be sending them through Tessen and therefore Segment. Product analytics folks don't need access to the custom events. You can read more about this [here](https://developers.google.com/analytics/devguides/collection/gajs/eventTrackerGuide).

### Option 2: Use `Tessen.track`
We can also send this through the `tessen.track()` method, but this would also mean we would send it to segment and NR browser (and we are already sending some of these events to NR browser, so to avoid duplication it might be best to option 1). 

## AC 
* [ ] Add GA tracking for `copyCodeBlock_click`
* [ ] Add GA tracking for `externalLink_click` 
* [ ] Check we are on alignment with tracking 

## Helpful links 
* Tracking events with google analytics:  https://developers.google.com/analytics/devguides/collection/gajs/eventTrackerGuide
* [instrumentation tracking spreadsheet](https://docs.google.com/spreadsheets/d/1VmgEp4n-GA3wvIjBy7rJQn3H7k_ytoqxvYG50X8xeAs/edit?usp=sharing)

caylahamann:
Figure out how to link the old attribute to the new one

tariqahmed:
Just a note from Sy Jones: 

Hello this topic has come up quite a few times because we’ve been doing an audit of all the site analytics
and it looks like DOCS is on their own Google analytics code that’s different from the rest of the website

caylahamann:
closing this because we do not need the custom events, everything is already in NR and all events needed by product analytics is automatically available to GA through tessen :)

PR #269, opened 2021-02-17
Add a useInstrumentedHandler hook

jerelmiller:
Adds a `useInstrumentedHandler` hook to easily augment handlers with `newrelic.addPageAction` calls.

### DIFF
--- a/packages/gatsby-theme-newrelic/src/hooks/useInstrumentedHandler.js
+++ b/packages/gatsby-theme-newrelic/src/hooks/useInstrumentedHandler.js
@@ -19,9 +19,7 @@ const useInstrumentedHandler = (handler, attributes) => {
           'You are attempting to instrument a handler, but the `actionName` property is not set. This will result in a no-op.'
         );
 
-        if (actionName) {
-          actionName && window.newrelic.addPageAction(actionName, attrs);
-        }
+        actionName && window.newrelic.addPageAction(actionName, attrs);
       }
 
       if (savedHandler.current) {
--- a/packages/gatsby-theme-newrelic/src/hooks/useInstrumentedHandler.js
+++ b/packages/gatsby-theme-newrelic/src/hooks/useInstrumentedHandler.js
@@ -1,0 +1,35 @@
+import { useCallback, useEffect, useRef } from 'react';
+import warning from 'warning';
+
+const useInstrumentedHandler = (handler, attributes) => {
+  const savedHandler = useRef();
+
+  useEffect(() => {
+    savedHandler.current = handler;
+  }, [handler]);
+
+  return useCallback(
+    (...args) => {
+      const { actionName, ...attrs } =
+        typeof attributes === 'function' ? attributes(...args) : attributes;
+
+      if (window.newrelic) {
+        warning(
+          actionName,
+          'You are attempting to instrument a handler, but the `actionName` property is not set. This will result in a no-op.'
+        );
+
+        if (actionName) {
+          actionName && window.newrelic.addPageAction(actionName, attrs);
+        }
+      }
+
+      if (savedHandler.current) {
+        return savedHandler.current(...args);
+      }
+    },
+    [attributes]
+  );
+};
+
+export default useInstrumentedHandler;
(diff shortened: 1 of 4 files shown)


packages/gatsby-theme-newrelic/src/hooks/useInstrumentedHandler.js
```diff
@@ -0,0 +1,35 @@
...
+  useEffect(() => {
+    savedHandler.current = handler;
+  }, [handler]);
+
+  return useCallback(
+    (...args) => {
+      const { actionName, ...attrs } =
+        typeof attributes === 'function' ? attributes(...args) : attributes;
+
+      if (window.newrelic) {
```
zstix, line 16: Do we need to double check that we have access to `window` here?

packages/gatsby-theme-newrelic/src/hooks/useInstrumentedHandler.js
```diff
@@ -0,0 +1,35 @@
...
+        warning(
+          actionName,
+          'You are attempting to instrument a handler, but the `actionName` property is not set. This will result in a no-op.'
+        );
+
+        if (actionName) {
+          actionName && window.newrelic.addPageAction(actionName, attrs);
```
zstix, line 23: Is the `actionName` necessary if we're already in the `if` block?

zstix (dismissed):
Looks good! Two minor questions

packages/gatsby-theme-newrelic/src/hooks/useInstrumentedHandler.js
jerelmiller: We shouldn't need to only because this function will only ever be called as part of a handler for a component, and will not be executed until the function is called.

Unless we come across a really unique use case, 99.99% of the uses of this hook will be something like the following:

```js
const handleClick = useInstrumentedHandler(() => console.log('clicked'), { actionName: 'click' })

<div onClick={handleClick} />
```

That `handleClick` doesn't get executed on a server render, so we should be safe.

packages/gatsby-theme-newrelic/src/hooks/useInstrumentedHandler.js
jerelmiller: Oops, I could have sworn I removed the `if`, but must have hit `u` by accident 😅 . Will remove the `if`

jerelmiller: Fixed in a752daf

packages/gatsby-theme-newrelic/src/hooks/useInstrumentedHandler.js
```diff
@@ -0,0 +1,33 @@
+import { useCallback, useEffect, useRef } from 'react';
+import warning from 'warning';
+
+const useInstrumentedHandler = (handler, attributes) => {
```
caylahamann, line 4: This doesn't need to be added in this PR, but I think this would be great to expand to handle all instrumentation. For example, we could have something like: 
```js
const useInstrumentedHandler = (handler, attributes, config) => {
```
where config could be an object with defaults `{newrelic: true, tessen: false, googleAnalytics: false}. And then we would have something like: 
```js
const instrumentationConfig =  {newrelic: false, tessen: true, googleAnalytics: true};

useInstrumentedHandler(() => console.log('clicked'), { actionName: 'click' }, instrumentationConfig);
```
Therefore, for all the tickets like #254 where we want to instrument to multiple tools we could just have one easy handler.

What do you think?

jerelmiller: Oooo that's a really good idea.

I think the trick here is figuring out how best to unify the attributes. NR Browser requires an `actionName`, Tessen requires both a `name` and `category` (which it then combines into a single string when it sends to New Relic), and Google Analytics [requires](https://developers.google.com/analytics/devguides/collection/analyticsjs/command-queue-reference#send) `eventCategory`, `eventAction`, `eventLabel`, and `eventValue`. 

If we can figure out how to best make sure that all of the required fields for each tracker are filled out, then I really like this idea.

jerelmiller:
Putting this PR on hold until we fully figure out the landscape for tracking based on the conversation the morning of 2/17

jerelmiller:
Since we decided to move forward with NR Browser tracking, this PR is unblocked

zstix (approved):

PR #270, opened 2021-02-17
Add custom attributes for newrelic option 

### LINKED ISSUE #256: [NR Browser] Set up addCustomAttribute for each page action
## Summary 
There are some properties that we want to record for each page action. Specifically, we want to attach to our subsequent `PageAction`s the locale the page is on, the path of the page, and the mode (dark or light) that the user is on. 

We can do this through the [`addCustomAttribute`](https://docs-preview.newrelic.com/docs/browser/new-relic-browser/browser-agent-spa-api/setcustomattribute-browser-agent-api) action in the browser agent. 

## AC 
* [ ] configure add custom attribute to fire before the window load event occurs 
* [ ] attach the path, locale, and mode to that event

caylahamann:
closes #256

### DIFF
--- a/packages/gatsby-theme-newrelic/gatsby-browser.js
+++ b/packages/gatsby-theme-newrelic/gatsby-browser.js
@@ -2,7 +2,6 @@ import React from 'react';
 import LayoutContext from './src/components/LayoutContext';
 import SplitIOProvider from './src/components/SplitIOProvider';
 import getSplitConfig from './src/utils/getSplitConfig';
-import getLocale from './src/utils/getLocale';
 
 export const wrapRootElement = ({ element }, pluginOptions) => {
   return (
@@ -18,25 +17,6 @@ export const wrapRootElement = ({ element }, pluginOptions) => {
   );
 };
 
-export const onClientEntry = (_, { newrelic, i18n }) => {
-  if (newrelic && window.newrelic) {
-    const mode = isDarkMode() ? 'dark' : 'light';
-    const locale = getLocale(i18n, window.location);
-
-    window.newrelic.setCustomAttribute('mode', mode);
-    window.newrelic.setCustomAttribute('locale', locale);
-  }
-};
-
-const isDarkMode = () => {
-  const localStorageTheme = localStorage.getItem('darkMode');
-
-  if (localStorageTheme) {
-    return JSON.parse(localeStorageTheme);
-  }
-
-  return document.body.classList.contains('dark-mode');
-};
-
+export { default as onClientEntry } from './gatsby/on-client-entry';
 export { default as onRouteUpdate } from './gatsby/on-route-update';
 export { default as wrapPageElement } from './gatsby/wrap-page-element';
--- a/packages/gatsby-theme-newrelic/gatsby-browser.js
+++ b/packages/gatsby-theme-newrelic/gatsby-browser.js
@@ -19,7 +19,7 @@ export const wrapRootElement = ({ element }, pluginOptions) => {
 };
 
 export const onClientEntry = (_, { newrelic, i18n }) => {
-  if (newrelic) {
+  if (newrelic && window.newrelic) {
     const mode = isDarkMode() ? 'dark' : 'light';
     const locale = getLocale(i18n, window.location);
 
--- a/packages/gatsby-theme-newrelic/gatsby-browser.js
+++ b/packages/gatsby-theme-newrelic/gatsby-browser.js
@@ -23,8 +23,8 @@ export const onClientEntry = (_, { newrelic, i18n }) => {
     const mode = isDarkMode() ? 'dark' : 'light';
     const locale = getLocale(i18n, window.location);
 
-    window.newrelic.addCustomAttribute('mode', mode);
-    window.newrelic.addCustomAttribute('locale', locale);
+    window.newrelic.setCustomAttribute('mode', mode);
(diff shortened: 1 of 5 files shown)


packages/gatsby-theme-newrelic/gatsby-browser.js
```diff
@@ -17,5 +18,25 @@ export const wrapRootElement = ({ element }, pluginOptions) => {
   );
 };
 
+export const onClientEntry = (_, { newrelic, i18n }) => {
+  if (newrelic && window.newrelic) {
```
jerelmiller, line 22: Out of curiosity, is this `newrelic` the theme's config value? If so, I'd be curious why we might depend on that to add custom attributes. Would `window.newrelic` be enough?

packages/gatsby-theme-newrelic/gatsby-browser.js
```diff
@@ -17,5 +18,25 @@ export const wrapRootElement = ({ element }, pluginOptions) => {
   );
 };
 
+export const onClientEntry = (_, { newrelic, i18n }) => {
```
jerelmiller, line 21: Out of curiosity, why did you choose `onClientEntry`? No wrong answer, just curious 😄 

[nit] To follow the pattern of the other hooks, could you also extract this to a `gatsby/on-client-entry.js` file? The only reason `wrapRootElement` is in here is because it differs in implementation from the `wrapRootElement` defined in `gatsby-ssr.js`

packages/gatsby-theme-newrelic/gatsby/on-route-update.js
```diff
@@ -1,7 +1,17 @@
 import trackViaTessen from '../src/utils/page-tracking/tessen';
+import getLocale from '../src/utils/getLocale';
 
 const onRouteUpdate = ({ location, prevLocation }, themeOptions) => {
   trackViaTessen({ location, prevLocation }, themeOptions);
+
+  if (themeOptions.newrelic && window.newrelic) {
+    const { i18n } = themeOptions;
```
jerelmiller, line 8:
```suggestion
    const { i18n: i18nConfig } = themeOptions;
```

Should this get the wrapped i18n config using the `withDefaults` helper? In the event `i18n` isn't configured, this value would be `undefined`, which may break the `getLocale` function. Also, could we name this `i18nConfig` just so we don't confuse it with the `i18n` instance? You can see an example here: https://github.com/newrelic/gatsby-theme-newrelic/blob/develop/packages/gatsby-theme-newrelic/gatsby/wrap-page-element.js#L15

In fact, it might even be better just to pass the full `themeOptions` to the `getLocale` function and have that function wrap using `withDefaults` so that you don't have to repeat it between the `onRouteUpdate` and `onClientEntry` hooks.

packages/gatsby-theme-newrelic/src/components/DarkModeToggle.js
```diff
@@ -15,6 +15,12 @@ const DarkModeToggle = ({ className, size, onClick }) => {
       onClick={(e) => {
         darkMode.toggle();
 
+        const mode = darkMode.value ? 'dark' : 'light';
+
+        if (window.newrelic) {
+          window.newrelic.setCustomAttribute('mode', mode);
```
jerelmiller, line 21:
```suggestion
        if (window.newrelic) {
          window.newrelic.setCustomAttribute('mode', darkMode.value ? 'dark' : 'light');
```

Very minor nit, we could save a couple lines be inlining the value here.

jerelmiller (commented):
Had a few very minor suggestions/questions for you!

packages/gatsby-theme-newrelic/gatsby-browser.js
caylahamann: Yes I could follow this pattern! and I chose `onClientEntry` it was the closest method to `window.onload` in the gatsby lifecycles hook. I could have also just done it all in `onRouterUpdate` (looking into it I really couldn't figure out a difference between those two other than the location properties that get passed in), so I could refactor this into `onRouterUpdate`.

packages/gatsby-theme-newrelic/gatsby/on-route-update.js
caylahamann: Good idea! I will make that change

packages/gatsby-theme-newrelic/gatsby-browser.js
caylahamann: Very true! `window.newrelic` should be enough.

packages/gatsby-theme-newrelic/gatsby/on-client-entry.js
```diff
@@ -0,0 +1,21 @@
+import getLocale from '../src/utils/getLocale';
+
+export const onClientEntry = (_, themeOptions) => {
```
jerelmiller, line 3: It looks like [you're looking for the `default` export](https://github.com/newrelic/gatsby-theme-newrelic/pull/270/files#diff-ae38556db84ac9f7923212ab9135c98208a4fdcd41f6764e43fe41641b57abbcR20), but here you're exporting this as a named export. Can you instead convert this to a default export?

```js
const onClientEntry = () => ...

export default onClientEntry
```

jerelmiller (commented):
One more minor thing, then I'll approve!

jerelmiller (approved):

PR #286, opened 2021-02-24
Mobile global nav updates

jerelmiller:
Closes https://github.com/newrelic/docs-website/issues/936

The global on a mobile view is a bit crammed with the list of sites. This PR makes some tweaks to the nav to ensure a nice mobile experience.

## Screenshots

IMG_3174
IMG_3175


zstix (approved):
First: _THANK YOU!_ This has been mildly annoying me for weeks.

Second: Is the "log in" icon clear? I wasn't sure what that icon originally represented (I had to look at the current site for it to make sense).

I approved it either way, could just be my brain not making the connection.

jerelmiller:
I spoke with @djsauble about this nav and he originally had the idea for the login icon. We can run with it and go back to the text if we find over time that this becomes confusing.

PR #295, opened 2021-03-01
Localizes paths for newrelic.com

caylahamann:
Localizes external link paths to newrelic.com

### DIFF
--- a/packages/gatsby-theme-newrelic/src/utils/localization.js
+++ b/packages/gatsby-theme-newrelic/src/utils/localization.js
@@ -17,10 +17,9 @@ export const localizeExternalLink = ({ link, locale }) => {
     return link;
   }
 
-  const paths = link.replace('https://', '').split('/');
+  const url = new URL(link);
 
-  paths.splice(1, 0, locale.locale).join('/');
+  url.pathname = localizePath({ path: url.pathname, locale });
 
-  return `https://${paths.join('/')}`;
+  return url.href;
 };
-//{ path: to, locale }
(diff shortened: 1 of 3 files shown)


packages/gatsby-theme-newrelic/src/utils/localization.js
```diff
@@ -11,3 +11,16 @@ export const localizePath = ({ path, locale }) => {
...
 };
+
+export const localizeExternalLink = ({ link, locale }) => {
+  if (locale.isDefault) {
+    return link;
+  }
+
+  const paths = link.replace('https://', '').split('/');
+
+  paths.splice(1, 0, locale.locale).join('/');
```
jerelmiller, line 22: Rather than using array methods, which get just a little hard to follow, you could clean this up a little bit by using the `URL` constructor. Doing so also allows you to take advantage of the `localizePath` helper above.

```js
const url = new URL(link)

url.pathname = localizePath({ path: url.pathname, locale })

return url.href
```

jerelmiller (commented):
Had one minor suggestion on constructing that pathname

packages/gatsby-theme-newrelic/src/utils/localization.js
caylahamann: oooh yes, I'm good with that! Let me make that refactor

caylahamann: @jerelmiller It has been updated! tested it out with all the links and works great :)

jerelmiller (approved):
Pulled it down and saw the sign up button went to the localized sign up page. Huzzah!

PR #299, opened 2021-03-02
Handles localized home link

caylahamann:
Handles the case of the localized link being the homepage

### DIFF
--- a/packages/gatsby-theme-newrelic/src/utils/localization.js
+++ b/packages/gatsby-theme-newrelic/src/utils/localization.js
@@ -5,10 +5,6 @@ export const localizePath = ({ path, locale }) => {
     return path;
   }
 
-  if (path === '/') {
-    return `/${locale.locale}/`;
-  }
-
   const [, base] = path.split('/');
 
   return base === locale.localizedPath
--- a/packages/gatsby-theme-newrelic/src/utils/localization.js
+++ b/packages/gatsby-theme-newrelic/src/utils/localization.js
@@ -5,6 +5,10 @@ export const localizePath = ({ path, locale }) => {
     return path;
   }
 
+  if (path === '/') {
+    return `/${locale.locale}/`;
+  }
+
   const [, base] = path.split('/');
 
   return base === locale.localizedPath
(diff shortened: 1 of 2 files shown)


packages/gatsby-theme-newrelic/src/utils/localization.js
```diff
@@ -5,6 +5,10 @@ export const localizePath = ({ path, locale }) => {
     return path;
   }
 
+  if (path === '/') {
```
jerelmiller, line 8: It seems this is already handled by the logic on line 14. I can spin up the theme on develop, switch to Japanese, and get the home link as I would expect. I also see this behavior on the current docs site if I switch to JP and click on the logo. Maybe I'm missing something?

caylahamann: Yeah that's what I thought so too, but I didn't understand why I was seeing the behavior I was seeing so I added an extra conditional. When you go directly to docs.newrelic.com/jp and click on the footer logo it takes you to the english homepage.

caylahamann: Actually I'm seeing it even when I switch to japanese while I'm on the homepage. it's only in the footer logo 🤔 . Maybe it's because we shadow that logo?

jerelmiller: Ah ha! Thats the ticket. It looks like we are importing `Link` from `gatsby`, not our custom `Link` component:

https://github.com/newrelic/gatsby-theme-newrelic/blob/develop/packages/gatsby-theme-newrelic/src/components/GlobalFooter.js#L6

Switch that import and I bet that fixes this issue.

jerelmiller (changes requested):
Let's use the feedback in the thread above as the change we need

packages/gatsby-theme-newrelic/src/utils/localization.js
caylahamann: sweeet! will make that change

caylahamann: all changed :)

jerelmiller (approved):
Looks like that fixed it! Thank you!

PR #305, opened 2021-03-03
Tracks tessen for sign up links and allows to specify additional props to the event

caylahamann:
tested it by console logging the event (because there was a no-op due to tessen not being configured for the demo site. Tracks when you click signup buttons, but not when you click a regular old external link. 

Also uses the external link component instead of `a` tag to separate actions for external links and links in general. 
Addresses https://github.com/newrelic/docs-website/issues/1113 once the theme is updated

### DIFF
--- a/packages/gatsby-theme-newrelic/src/components/ExternalLink.js
+++ b/packages/gatsby-theme-newrelic/src/components/ExternalLink.js
@@ -60,7 +60,7 @@ ExternalLink.propTypes = {
   children: PropTypes.node,
   href: PropTypes.string,
   onClick: PropTypes.func,
-  trackingProps: PropTypes.oneOfType([PropTypes.object, PropTypes.string]),
+  instrumentation: PropTypes.oneOfType([PropTypes.object, PropTypes.string]),
 };
 
 export default ExternalLink;
--- a/packages/gatsby-theme-newrelic/src/components/ExternalLink.js
+++ b/packages/gatsby-theme-newrelic/src/components/ExternalLink.js
@@ -9,14 +9,25 @@ import { useLocation } from '@reach/router';
 const isNewRelic = (to) => to.startsWith('https://newrelic.com');
 const isSignup = (to) => to.startsWith('https://newrelic.com/signup');
 
-const ExternalLink = ({ children, onClick, href, component, ...props }) => {
+const ExternalLink = ({
+  children,
+  onClick,
+  href,
+  trackingProps = {},
+  ...props
+}) => {
   const locale = useLocale();
   const tessen = useTessen();
   const location = useLocation();
+  const properties =
+    typeof trackingProps === 'string'
+      ? { customProp: trackingProps }
+      : trackingProps;
 
   const handleClick = useInstrumentedHandler(onClick, {
     actionName: 'externalLink_click',
     href: href,
+    ...properties,
   });
 
   const link = isNewRelic(href)
@@ -28,7 +39,7 @@ const ExternalLink = ({ children, onClick, href, component, ...props }) => {
     tessen.track('stitchedPathLinkClick', 'DocPageLinkClick', {
       href: link,
       path: location.pathname,
-      component,
+      ...properties,
     });
   };
 
@@ -49,6 +60,7 @@ ExternalLink.propTypes = {
   children: PropTypes.node,
   href: PropTypes.string,
   onClick: PropTypes.func,
+  trackingProps: PropTypes.oneOfType([PropTypes.object, PropTypes.string]),
 };
 
 export default ExternalLink;
--- a/packages/gatsby-theme-newrelic/src/components/ExternalLink.js
+++ b/packages/gatsby-theme-newrelic/src/components/ExternalLink.js
@@ -3,26 +3,40 @@ import PropTypes from 'prop-types';
 import useInstrumentedHandler from '../hooks/useInstrumentedHandler';
 import { localizeExternalLink } from '../utils/localization';
 import useLocale from '../hooks/useLocale';
+import useTessen from '../hooks/useTessen';
+import { useLocation } from '@reach/router';
 
 const isNewRelic = (to) => to.startsWith('https://newrelic.com');
+const isSignup = (to) => to.startsWith('https://newrelic.com/signup');
 
-const ExternalLink = ({ children, onClick, href, ...props }) => {
+const ExternalLink = ({ children, onClick, href, component, ...props }) => {
   const locale = useLocale();
+  const tessen = useTessen();
+  const location = useLocation();
+
+  const handleClick = useInstrumentedHandler(onClick, {
+    actionName: 'externalLink_click',
+    href: href,
+  });
 
   const link = isNewRelic(href)
     ? localizeExternalLink({ link: href, locale })
     : href;
 
-  const handleClick = useInstrumentedHandler(onClick, {
-    actionName: 'externalLink_click',
-    href: link,
-  });
+  const trackSignUp = () => {
+    handleClick();
+    tessen.track('stitchedPathLinkClick', 'DocPageLinkClick', {
+      href: link,
+      path: location.pathname,
+      component,
+    });
+  };
 
   return (
     <a
       {...props}
       href={link}
-      onClick={handleClick}
+      onClick={isSignup(href) ? trackSignUp : handleClick}
       target="_blank"
       rel="noopener noreferrer"
     >
--- a/packages/gatsby-theme-newrelic/src/components/ExternalLink.js
+++ b/packages/gatsby-theme-newrelic/src/components/ExternalLink.js
@@ -34,7 +34,7 @@ const ExternalLink = ({
     ? localizeExternalLink({ link: href, locale })
     : href;
 
-  const trackSignUp = () => {
+  const trackSignUp = (event) => {
     handleClick();
     tessen.track('stitchedPathLinkClick', 'DocPageLinkClick', {
       href: link,
--- a/packages/gatsby-theme-newrelic/src/components/ExternalLink.js
+++ b/packages/gatsby-theme-newrelic/src/components/ExternalLink.js
@@ -13,16 +13,16 @@ const ExternalLink = ({
   children,
   onClick,
   href,
-  trackingProps = {},
+  instrumentation = {},
   ...props
 }) => {
   const locale = useLocale();
   const tessen = useTessen();
   const location = useLocation();
   const properties =
-    typeof trackingProps === 'string'
-      ? { customProp: trackingProps }
-      : trackingProps;
+    typeof instrumentation === 'string'
+      ? { customProp: instrumentation }
+      : instrumentation;
 
   const handleClick = useInstrumentedHandler(onClick, {
     actionName: 'externalLink_click',
--- a/packages/gatsby-theme-newrelic/src/components/ExternalLink.js
+++ b/packages/gatsby-theme-newrelic/src/components/ExternalLink.js
@@ -1,66 +1,13 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import useInstrumentedHandler from '../hooks/useInstrumentedHandler';
-import { localizeExternalLink } from '../utils/localization';
-import useLocale from '../hooks/useLocale';
-import useTessen from '../hooks/useTessen';
-import { useLocation } from '@reach/router';
+import Link from './Link';
 
-const isNewRelic = (to) => to.startsWith('https://newrelic.com');
-const isSignup = (to) => to.startsWith('https://newrelic.com/signup');
-
-const ExternalLink = ({
-  children,
-  onClick,
-  href,
-  instrumentation = {},
-  ...props
-}) => {
-  const locale = useLocale();
-  const tessen = useTessen();
-  const location = useLocation();
-  const properties =
-    typeof instrumentation === 'string'
-      ? { customProp: instrumentation }
-      : instrumentation;
-
-  const handleClick = useInstrumentedHandler(onClick, {
-    actionName: 'externalLink_click',
-    href: href,
-    ...properties,
-  });
-
-  const link = isNewRelic(href)
-    ? localizeExternalLink({ link: href, locale })
-    : href;
-
-  const trackSignUp = (event) => {
-    handleClick();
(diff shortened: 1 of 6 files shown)


packages/gatsby-theme-newrelic/src/components/Link.js
```diff
@@ -40,18 +35,7 @@ const Link = ({ to, onClick, ...props }) => {
...
-    return (
-      <a
-        href={link}
-        onClick={handleExternalLinkClick}
-        target="_blank"
-        rel="noopener noreferrer"
-        {...props}
-      />
-    );
+    return <ExternalLink href={to} onClick={onClick} {...props} />;
```
jerelmiller, line 38: Could we instead do the reverse? I would like to deprecate `ExternalLink` as its got the same functionality as `Link` (as you've already seen). Ideally everything goes through `Link` so that we don't need 2 components. At the very least, could you revert this change and apply the same logic here? I'm ok with duplicate logic in the short term.

packages/gatsby-theme-newrelic/src/components/ExternalLink.js
```diff
@@ -3,26 +3,40 @@ import PropTypes from 'prop-types';
...
 import { localizeExternalLink } from '../utils/localization';
 import useLocale from '../hooks/useLocale';
+import useTessen from '../hooks/useTessen';
+import { useLocation } from '@reach/router';
 
 const isNewRelic = (to) => to.startsWith('https://newrelic.com');
+const isSignup = (to) => to.startsWith('https://newrelic.com/signup');
 
-const ExternalLink = ({ children, onClick, href, ...props }) => {
+const ExternalLink = ({ children, onClick, href, component, ...props }) => {
```
jerelmiller, line 12: Rather than a `component` prop, could you instead make this an `instrumentation` prop to match some of our other components like the [`CreateIssueButton`](https://github.com/newrelic/gatsby-theme-newrelic/blob/develop/packages/gatsby-theme-newrelic/src/components/CreateIssueButton.js#L61-L63)? Using `instrumentation` better calls out the purpose of the prop so that is clear `component` is used for instrumentation purposes. Without it, this appears to change the component that is used to render the link itself:

```
<ExternalLink component="Something" /> // <-- looks like "Something" is supposed to be the component rendered here
```

packages/gatsby-theme-newrelic/src/components/ExternalLink.js
```diff
@@ -3,26 +3,40 @@ import PropTypes from 'prop-types';
...
   const link = isNewRelic(href)
     ? localizeExternalLink({ link: href, locale })
     : href;
 
-  const handleClick = useInstrumentedHandler(onClick, {
-    actionName: 'externalLink_click',
-    href: link,
-  });
+  const trackSignUp = () => {
+    handleClick();
```
jerelmiller, line 27: [nit] Can you pass through the `event` to this click handler? While most of the time we don't use it, we should forward it in case something does need to use it (like a `preventDefault` or something).

jerelmiller (changes requested):

packages/gatsby-theme-newrelic/src/components/ExternalLink.js
caylahamann: Yeah I was just changing this haha, which is why I hadn't posted it yet! I was making it `trackingProps` that could either be a string or an object of additional properties.

packages/gatsby-theme-newrelic/src/components/Link.js
caylahamann: Could I potentially create a Link folder and then a file with another ExternalLink component or some other name? everything would be served out of `Link`, but it's just so messy with it all being in one file

packages/gatsby-theme-newrelic/src/components/ExternalLink.js
caylahamann: yep!

packages/gatsby-theme-newrelic/src/components/Link.js
jerelmiller: Maybe we should just let `ExternalLink` use `Link`. That might be the best of both worlds until we deprecate everything.

```js
const ExternalLink = ({ href, ...props }) => {
  return <Link to={href} {...props} />
}
```

Thoughts?

packages/gatsby-theme-newrelic/src/components/ExternalLink.js
jerelmiller: Could we standardize on `instrumentation`? If we go with `trackingProps` could we then update the other components so we can settle on a pattern for this? That would be super helpful if we could.

caylahamann: it has been now changed to `instrumentation`

packages/gatsby-theme-newrelic/src/components/Link.js
caylahamann: I'm okay with that! My issue was more towards that `Link` becomes a super dense component - I pushed up the changes to show what I mean! I'm okay with doing it this way, I just wanted to separate out the concerns of external links and non-external links :)

jerelmiller: Got it! Sorry took me a bit to understand what you meant (too much going on in my brain right now 😬 ). 


I'm ok with this for now. I do agree that maybe we could look at refactoring this a bit, but for now, lets leave it as-is. I don't _think_ we'll need too much more here, but would be good to monitor that over time.

packages/gatsby-theme-newrelic/src/components/ExternalLink.js
```diff
@@ -1,40 +1,13 @@
...
-      href={link}
-      onClick={handleClick}
-      target="_blank"
-      rel="noopener noreferrer"
-    >
-      {children}
-    </a>
-  );
+const ExternalLink = ({ href, ...props }) => {
+  return <Link to={href} {...props} />;
```
jerelmiller, line 6: 🎉

packages/gatsby-theme-newrelic/src/components/Link.js
```diff
@@ -31,6 +32,29 @@ const Link = ({ to, onClick, ...props }) => {
     }
   `);
 
+  const properties =
+    typeof instrumentation === 'string'
+      ? { customProp: instrumentation }
```
jerelmiller, line 37: Looking at this more, I'm not so sure I like this. The `customProp` name feels a bit arbitrary, and not super useful. For example, if you defined this as `<Link instrumentation="GlobalHeader" />`, what would you expect? By allowing a string, it's not very clear what the intent is here.

I think we should enforce an object so that the additional instrumented properties are very clearly defined. Let's not allow a string as it could lead to more confusion than its worth.

packages/gatsby-theme-newrelic/src/components/Link.js
```diff
@@ -31,6 +32,29 @@ const Link = ({ to, onClick, ...props }) => {
...
+      : instrumentation;
+
+  const handleClick = useInstrumentedHandler(onClick, {
+    actionName: 'externalLink_click',
+    href: to,
+    ...properties,
```
jerelmiller, line 43: Since `Link` is used everywhere, most cases are not going to have an `instrumentation` prop defined. I _believe_ this would blow up when `instrumentation` is `undefined`.  Could you make sure this defaults to an empty object if its `undefined`?

packages/gatsby-theme-newrelic/src/components/Link.js
```diff
@@ -31,6 +32,29 @@ const Link = ({ to, onClick, ...props }) => {
     }
   `);
 
+  const properties =
+    typeof instrumentation === 'string'
+      ? { customProp: instrumentation }
+      : instrumentation;
+
+  const handleClick = useInstrumentedHandler(onClick, {
```
jerelmiller, line 40: [nit] This used to be called `handleExternalLinkClick` to clearly call out that this handler was meant for the external link case only. Could you rename to be the same?

packages/gatsby-theme-newrelic/src/components/Link.js
```diff
@@ -31,6 +32,29 @@ const Link = ({ to, onClick, ...props }) => {
...
+    actionName: 'externalLink_click',
+    href: to,
+    ...properties,
+  });
+
+  const link = isNewRelic(to) ? localizeExternalLink({ link: to, locale }) : to;
+
+  const trackSignUp = (event) => {
+    handleClick();
+    tessen &&
```
jerelmiller, line 50: Since you're pulling `tessen` from `useTessen`, no need to check if its defined here. `tessen` will always be defined, even if its not installed on the app. The [`createTessen` helper](https://github.com/newrelic/gatsby-theme-newrelic/blob/develop/packages/gatsby-theme-newrelic/src/utils/createTessen.js) checks to make sure `window.Tessen` is available and will result in a no-op if its not defined. Since `tessen` will always be defined here, the `track` will always get evaluated. Could you remove this since its unnecessary?

jerelmiller (changes requested):

packages/gatsby-theme-newrelic/src/components/Link.js
caylahamann: I was more thinking for the purposes of content contributors wanted to add instrumentation, which is why i thought maybe a string might be easier to create rather than an object at times! 

I agree it's definitely arbitrary, but one of the requests is that sign up links could be tracked wherever they occur and most of the time that should be configured by docs writers. I'm totally for enforcing the object, just will have to make it clear to content contributors that this is the way we would like it configured! :)

packages/gatsby-theme-newrelic/src/components/Link.js
caylahamann: instrumentation is already defaulted to an empty object!! if you scroll up you can see that :)

jerelmiller: 🤦‍♂️ Wow. I need to go outside and take a walk. Thanks!!

packages/gatsby-theme-newrelic/src/components/Link.js
caylahamann: yep! I added it in because one of the tests was resulting in a `tessen is not defined`

packages/gatsby-theme-newrelic/src/components/Link.js
caylahamann: Oops yes, sorry I was just pasting in from when I had it namespaced to ExternalLink! it has been changed back

packages/gatsby-theme-newrelic/src/components/Link.js
jerelmiller: Absolutely! I think its reasonable to ask that they define this as an object so that properties are named accordingly. I just think `customProp` is going to get lost, or is meaningless in this case. Let's do what we can to teach content contributors how to add instrumentation when needed 🙂

packages/gatsby-theme-newrelic/src/components/GlobalHeader.js
```diff
@@ -51,11 +51,6 @@ const GlobalHeader = ({ className }) => {
   const [searchQuery, setSearchQuery] = useState('');
   const { t } = useThemeTranslation();
 
-  const handleSignupClick = useInstrumentedHandler(null, {
-    actionName: 'signup_click',
```
jerelmiller, line 55: One more question while I'm here. Do we not want to instrument `signup_click` through NR browser anymore?

jerelmiller:
The recent updates look good to me! I see some tests failing, so I'll wait to approve until after those are fixed.

packages/gatsby-theme-newrelic/src/components/GlobalHeader.js
caylahamann: no we do not! that was not part of the plan - it was always supposed to be sent through tessen

caylahamann: also the `signup_clicks` are actually meant to be named `stitchedDocPath` and by using the link component we are covering this case already

packages/gatsby-theme-newrelic/src/components/Link.js
```diff
@@ -60,6 +75,7 @@ const Link = ({ to, onClick, ...props }) => {
 Link.propTypes = {
   onClick: PropTypes.func,
   to: PropTypes.string.isRequired,
+  instrumentation: PropTypes.object),
```
jerelmiller, line 78:
```suggestion
  instrumentation: PropTypes.object,
```

Looks like the build/tests are failing due to this syntax error 😄

caylahamann: I really need to get some sleep or something 😦 sorry about that, very careless

jerelmiller (approved):
Thanks so much for making those changes!

jpvajda:
Great work on this, scanning the changes I was curious did we change anything in our event naming that should be reflected on the wiki page: https://github.com/newrelic/docs-website/wiki/Event-Data--and--Instrumentation @caylahamann

caylahamann:
@jpvajda No we did not! Actually it now reflects what is in that wiki, before the wiki was slightly inaccurate about the tessen events. the `signup_link` was never supposed to be tracked as just an NR event - it was supposed to be `TDOC_DocPageLinkClick_stitchedPathLinkClick` and be sent to Tessen. With this change, we are now up to parity with instrumentation (plus some) with the old docs site and all of that is reflected in the wiki.

jpvajda:
@caylahamann ok thanks! do we need to add some context to this event then? 

https://github.com/newrelic/docs-website/wiki/Event-Data--and--Instrumentation#event-data-1


TDOC_DocPageLinkClick_stitchedPathLinkClick | occurs when a user clicks on a button deemed a "stitched path"
-- | --

We don't explictely callout `signup_link` or clicks on a signup buttons, so I'm wonder if we just need a little context about that.

jpvajda:
@caylahamann any thoughts to my question?

caylahamann:
@jpvajda I think that might be more of a question for the product analytics folks. We might be missing some cases here - if a stitchedPathLink is just a sign up link or some other link that leads to eventual sign up? It's not entirely clear to me. We right now just track sign up links.

PR #306, opened 2021-03-03
fix(nav): on search the item collapse should work

AubreyHewes:
fixes https://github.com/newrelic/docs-website/issues/938

cases (demo app):

 * Nav search for "relic" displays all items containing "relic" and the collapse works
   * Then search "data" displays all items containing "data" (previous collapsed are no longer collapsed) and the collapse works


jerelmiller (approved):
So simple! Thanks so much for getting this fixed!

PR #307, opened 2021-03-04
fix(global-search): when cancelling a search the search input is cleared

AubreyHewes:
fixes https://github.com/newrelic/docs-website/issues/937

Type "test" in the global search
Click the "x"... global search is cleared (previously was not)


jerelmiller (approved):
D'oh! Caught by the z-index. Nice catch!

AubreyHewes:
Took me a while to find the place... but that was due to not knowing the code (the demo app helps a great deal), the nav-search worked, so why not the global search? well the fast access icon ;-) ...

PR #308, opened 2021-03-04
fix(search): invalid css

AubreyHewes:
While solving a few other search-component issues I noticed that the search component had invalid css;

#### How did I notice this?

 * Well the IDE said it is invalid css.. and I know that it is.

#### How was this never noticed before?

 * Bad CI linting
 * Someone editing in notepad?

#### How could this ever have happened?

 * Someone editing in notepad?


jerelmiller (approved):
Ha, good catch as well! All of us _should_ have ESLint enabled, so weird that we didn't catch this before. Glad you saw it!

PR #311, opened 2021-03-05
chore: new labels

jpvajda:
Closes https://github.com/newrelic/docs-website/issues/1096

This PR adds the preferred labels to the create issue button component

### DIFF
--- a/packages/gatsby-theme-newrelic/src/components/CreateIssueButton.js
+++ b/packages/gatsby-theme-newrelic/src/components/CreateIssueButton.js
@@ -5,6 +5,7 @@ import GitHubIssueButton from './GitHubIssueButton';
 import Icon from './Icon';
 import useThemeTranslation from '../hooks/useThemeTranslation';
 import useInstrumentedHandler from '../hooks/useInstrumentedHandler';
+import { graphql, useStaticQuery } from 'gatsby';
 
 const ISSUE_BODY = `
 <!-- Thanks for filing an issue on our docs! Your feedback helps us improve our
@@ -38,12 +39,32 @@ const CreateIssueButton = ({
     component: instrumentation?.component,
   });
 
+  const { site } = useStaticQuery(graphql`
+    query CreateIssueQuery {
+      site {
+        siteMetadata {
+          siteUrl
+        }
+      }
+    }
+  `);
+
+  const {
+    siteMetadata: { siteUrl },
+  } = site;
+
+  const inDocSite = [
+    'https://docs.newrelic.com',
+    'https://developer.newrelic.com',
+    'https://opensource.newrelic.com',
+  ].includes(siteUrl);
+
   return (
     <GitHubIssueButton
       {...props}
       issueTitle={pageTitle && `Issue: ${pageTitle}`}
       issueBody={ISSUE_BODY}
-      labels={['feedback', 'feedback-issue']}
+      labels={inDocSite ? ['feedback', 'feedback-issue'] : 'bug'}
       onClick={handleClick}
     >
       <Icon
--- a/packages/gatsby-theme-newrelic/src/components/CreateIssueButton.js
+++ b/packages/gatsby-theme-newrelic/src/components/CreateIssueButton.js
@@ -27,7 +27,7 @@ your environment (operating system, application framework, etc.).
 `;
 
 const CreateIssueButton = ({
-  labels = ['feedback', 'feedback-issue', 'bug'],
+  labels = ['bug'],
   instrumentation,
   pageTitle,
   onClick,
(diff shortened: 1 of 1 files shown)


packages/gatsby-theme-newrelic/src/components/CreateIssueButton.js
```diff
@@ -43,7 +43,7 @@ const CreateIssueButton = ({
       {...props}
       issueTitle={pageTitle && `Issue: ${pageTitle}`}
       issueBody={ISSUE_BODY}
-      labels={['bug']}
+      labels={['feedback', 'feedback-issue']}
```
jerelmiller, line 46: Just to verify, do we want this to apply to all sites? Or just the docs site?

jerelmiller (approved):
The change looks great! I had one question for you just to make sure we are updating this in the way we think we are.

packages/gatsby-theme-newrelic/src/components/CreateIssueButton.js
jpvajda: @jerelmiller ideally just docs site, as these  labels aren't used on developer / oss for tracking issues.  so if there is some IF / ELSE logic I can add here, I'm happy to do that.

jerelmiller: Ok! In that case, I think we should convert this into a prop for this component so that each site can specify its own set of labels. I think it probably still makes sense to default to `['bug']` if the prop isn't defined. Here are the changes I would make:

```js
const CreateIssueButton = ({ labels = ['bug'] /* <-- add `labels` prop }) => {
  // ...
  return (
    <GitHubIssueButton
      // ...
      labels={labels} // <--- use props instead of hard-coding the value here
    />
}

CreateIssueButton.propTypes = {
  labels: PropTypes.arrayOf(PropTypes.string) // <--- add prop type for new labels
}
```

We'll want to update the README as well to list the new `labels` prop for this component.

I'm going to dismiss my review until this is in place so we can get this working.

jerelmiller (changes requested):
Since we've identified that we want each site to be able to override the labels, I'm going to request those changes before approving this PR.

jpvajda:
@jerelmiller cool it's on m `todo` list for today.

packages/gatsby-theme-newrelic/src/components/CreateIssueButton.js
```diff
@@ -38,12 +39,32 @@ const CreateIssueButton = ({
     component: instrumentation?.component,
   });
 
+  const { site } = useStaticQuery(graphql`
```
jerelmiller, line 42: Rather than using a static query here, could we instead just make the `labels` a prop to this component?

jpvajda: sure... though I may need a pointer on how do that approach. 😄

jerelmiller: See [this comment](https://github.com/newrelic/gatsby-theme-newrelic/pull/311#discussion_r587991463). This should provide the instructions you need 🙂

jpvajda: oh I didn't even see that... ok yeah let me change the approach. at least I got practice with using static queries.

jpvajda:
@jerelmiller so next steps are: once this approved / merged we just set the prop type we want on each site?

jerelmiller:
Thats correct!

packages/gatsby-theme-newrelic/src/components/CreateIssueButton.js
```diff
@@ -27,6 +27,7 @@ your environment (operating system, application framework, etc.).
 `;
 
 const CreateIssueButton = ({
+  labels = ['feedback', 'feedback-issue', 'bug'],
```
jerelmiller, line 30:
```suggestion
  labels = ['bug'],
```

Last minor thing! This is the default value, which will kick in if the `label` prop is not provided when using the `CreateIssueButton` on a site. Since `feedback` and `feedback-issue` are docs specific, let's default this to `bug` for all other sites. We'll override this default on the docs site when this change is merged in. Other than this, the change looks great!

jerelmiller (approved):

jerelmiller:
Thanks for updating!

jpvajda:
@jerelmiller also, the other sites will need to be updated this theme version as well to use this new proptype.

jerelmiller:
> use this new proptype

Yes and no 😄 . That default value (the `= ['bug']`) you have defined will be the default set of labels on any site that doesn't provide a `labels` prop to this component. This should maintain backwards compatibility with the other sites.

PR #315, opened 2021-03-06
Initializes tessen only once

caylahamann:
We were not recording updates on our pages because we were initialize tessen a bunch of times. As a result, we were initializing Segment and NR a bunch of times as well, which let to downstream errors that prevented us from recording our data. This should ensure that we only call initialize once!

### DIFF
--- a/packages/gatsby-theme-newrelic/src/utils/page-tracking/tessen.js
+++ b/packages/gatsby-theme-newrelic/src/utils/page-tracking/tessen.js
@@ -82,8 +82,6 @@ const initializeTessenTracking = ({ config, env, location }) => (
       },
     });
 
-    window.Tessen.debugLevel(2);
-
     window.Tessen.identify({});
 
     options.trackPageView && trackPageView({ config, env, location });
--- a/packages/gatsby-theme-newrelic/src/utils/page-tracking/tessen.js
+++ b/packages/gatsby-theme-newrelic/src/utils/page-tracking/tessen.js
@@ -2,6 +2,7 @@ import createTessen from '../createTessen';
 import warning from 'warning';
 import { getResolvedEnv, getTessenConfig } from '../config';
 import { canTrack } from '../tracking';
+import { once } from 'lodash';
 
 const warnAboutNoop = (pageView) => {
   warning(
@@ -26,11 +27,17 @@ const trackViaTessen = ({ location }, themeOptions) => {
     return;
   }
 
-  window.initializeTessenTracking = initializeTessenTracking({
-    config: tessenConfig,
-    env,
-    location,
-  });
+  window.initializeTessenTracking = once(
+    initializeTessenTracking({
+      config: tessenConfig,
+      env,
+      location,
+    })
+  );
+
+  if (!canTrack()) {
+    return;
+  }
 
   window.initializeTessenTracking();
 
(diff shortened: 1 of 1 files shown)


packages/gatsby-theme-newrelic/src/utils/page-tracking/tessen.js
```diff
@@ -26,11 +27,17 @@ const trackViaTessen = ({ location }, themeOptions) => {
     return;
   }
 
-  window.initializeTessenTracking = initializeTessenTracking({
-    config: tessenConfig,
-    env,
-    location,
-  });
+  window.initializeTessenTracking = once(
```
jerelmiller, line 30: So unfortunately because `trackViaTessen` is called each time the `onRouteUpdate` hook is run, `once` here doesn't quite work as expected. While this will prevent multiple `initializeTessenTracking()` calls in the same route update, the next time a user navigates to another page, this function gets reinitialized.

In order to ensure this will only run once ever, you'll need a stable reference to the function wrapped by `once` to ensure it doesn't get redefined. The only way to really do that here is to wrap the function on line 73 with `once` to ensure that function is only called one time. The only problem with that, however, is you may run into closure issues with the `location` variable since that variable could become stale over time if the initialize function isn't called right away.

There are a couple ways I can think get fix both of these scenarios:

1) Make the `location` an argument to `window.initializeTessenTracking` so that the caller can send the current value. This is something I believe you had proposed earlier. This would eliminate the closure issue and allow you to wrap the function below with `once`

2) Use a "global" variable as a flag to track when the value has been initialized, and return early if its set. What do I mean by this?

```js
let initialized = false

window.initializeTessen = initializeTessenTracking({ location, env, config })

// ...

const initializeTessenTracking = (...) => (...) => {
  if (initialized) {
    return
  }

  window.Tessen.load(...)
  initialized = true
}
```

I generally shy away from the 2nd option because I hate mutable values, but that technique does work well here.

Just wanted to throw this out there as a something I noticed. That `once` is partially working as expected, but it would be great to get that to work for the lifecycle of the page until the site is reloaded, not just for the route.

caylahamann: question: what is the difference between `route` updates and navigating to another page?

jerelmiller: They are synonymous (sorry for the confusion). The `onRouteUpdate` fires both when the page is first loaded and when a user user navigates to another page, either by clicking on a link, or the app redirects them to another page. Essentially any time the URL changes, this hook fires.

caylahamann: I checked this case and I don't _think_ that happens. Maybe I'm missing a case, but `once` executes around the actual initialization, so while `window.initializeTessenTracking` can get called multiple times, the underlying initialization gets only called once. When `trackViaTessen` gets called the underlying function initialization function (`initializeTessenTracking`) will have then been assigned to null and the values (including location) will have been bound, right? 

I tested this with the debugger and console.log statements in the initialization function and there was no other initializations except for on the first page load or when clicking the cookieconsent! I also could be not understanding the edge cases here, but I think we have this covered

caylahamann: OH I might understand, if we're calling `trackViaTessen` will we have instantiated `initializeTessenTracking` again and then the once function will no longer have been called? Like each time we call `window.initializeTessenTracking` will we have a new instance of that function and there the `once` no longer applies?

caylahamann: Still then confused about why I didn't see any other initializations when playing around with the demo site and navigating to other pages.... I'm going to double check

caylahamann:
@jerelmiller decided to use an initialization variable! when I did option 1 it wasn't getting called at all except if I did `window.initializeTessenTracking(location)()` and the result of that was it getting called multiple times. I don't know why I was seeing no multiple tracking initializations in the other way I did it, because when I pulled it down I no longer saw that 🤔 but this way works, checked it out with the debugLevel function!

jerelmiller (approved):
Looks great! Thanks so much for making this change

PR #321, opened 2021-03-10
Add swiftype type field for all NR subdomains

roadlittledawn:
Adds some logic to `<SEO />` component to get New Relic subdomain value from `siteMetaData.siteUrl` and the locale of the page if it's not the default (so any language other than English) and set the swiftype `type` field. Those values are currently set to the subdomain value in the Swiftype index.

Does not generate the meta tag if not on a New Relic subdomain (ex: localhost)

This requires opensource, docs, and developer sites to remove the code they use to generate the `type` field.

### Example meta tags
Docs site (English)

     <meta class="swiftype" name="type" data-type="enum" content="docs">

Docs site (Japanese)

    <meta class="swiftype" name="type" data-type="enum" content="docs-jp">

## Related issues
- https://github.com/newrelic/docs-website/issues/1204

### DIFF
--- a/packages/gatsby-theme-newrelic/src/components/SEO.js
+++ b/packages/gatsby-theme-newrelic/src/components/SEO.js
@@ -41,7 +41,7 @@ const SEO = ({ title, location, children }) => {
   const getSwiftypeSiteType = () => {
     const nrSubDomain = /.*\.newrelic\.com/.test(location.hostname)
       ? location.hostname.split('.')[0]
-      : 'demo';
+      : null;
     const localeString = locale.isDefault ? '' : `-${locale.locale}`;
     return nrSubDomain ? nrSubDomain.concat(localeString) : null;
   };
--- a/packages/gatsby-theme-newrelic/src/components/SEO.js
+++ b/packages/gatsby-theme-newrelic/src/components/SEO.js
@@ -67,7 +67,11 @@ const SEO = ({ title, location, children }) => {
         );
       })}
       {getSwiftypeSiteType() && (
-        <meta class="swiftype" name="type" content={getSwiftypeSiteType()} />
+        <meta
+          className="swiftype"
+          name="type"
+          content={getSwiftypeSiteType()}
+        />
       )}
       {children}
     </Helmet>
@@ -78,6 +82,7 @@ SEO.propTypes = {
   title: PropTypes.string,
   location: PropTypes.shape({
     pathname: PropTypes.string,
+    hostname: PropTypes.string,
   }).isRequired,
   children: PropTypes.node,
 };
--- a/packages/gatsby-theme-newrelic/src/components/SEO.js
+++ b/packages/gatsby-theme-newrelic/src/components/SEO.js
@@ -38,6 +38,14 @@ const SEO = ({ title, location, children }) => {
       ? location.pathname
       : location.pathname.replace(new RegExp(`^\\/${locale.locale}`), '/');
 
+  const getSwiftypeSiteType = () => {
+    const nrSubDomain = /.*\.newrelic\.com/.test(location.hostname)
+      ? location.hostname.split('.')[0]
+      : null;
+    const localeString = !locale.isDefault ? `-${locale.locale}` : '';
+    return nrSubDomain ? nrSubDomain.concat(localeString) : null;
+  };
+
   return (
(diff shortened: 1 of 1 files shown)


zstix (approved):

packages/gatsby-theme-newrelic/src/components/SEO.js
```diff
@@ -38,6 +38,14 @@ const SEO = ({ title, location, children }) => {
       ? location.pathname
       : location.pathname.replace(new RegExp(`^\\/${locale.locale}`), '/');
 
+  const getSwiftypeSiteType = () => {
+    const nrSubDomain = /.*\.newrelic\.com/.test(location.hostname)
```
jerelmiller, line 42: I worry that `hostname` is not available on a build since it doesn't execute in a browser. `pathname` makes sense since Gatsby knows which page it is currently building. Can we verify this works as expected when the site is built and not just at runtime?

packages/gatsby-theme-newrelic/src/components/SEO.js
```diff
@@ -38,6 +38,14 @@ const SEO = ({ title, location, children }) => {
...
+      ? location.hostname.split('.')[0]
+      : 'demo';
```
jerelmiller, line 44: I'm not a huge fan of this fallback. While this is ok for the demo site itself, this is a very strange value for any other site. Could we have a bit more sensible default?

packages/gatsby-theme-newrelic/src/components/SEO.js
```diff
@@ -38,6 +38,14 @@ const SEO = ({ title, location, children }) => {
...
+    const localeString = !locale.isDefault ? `-${locale.locale}` : '';
```
jerelmiller, line 45: [nit] I generally like to use positive connotation first. Typically its easier to think in the positive and helps your brain digest it faster (i.e. "if this... then that" vs "if not this... then that"). Would you be willing to swap these?

Feel free to leave it as-is if you feel strongly. Figured I'd point this out as its really helped me over the years understand conditional logic much quicker when reading through a codebase.

packages/gatsby-theme-newrelic/src/components/SEO.js
roadlittledawn: yeah i was actually waffling between making it `null` if not running on a NR subdomain, which would remove the `<meta>` tag if kept as is. which is maybe fine? OR we could just let use localhost, but that feels a little weird too.

packages/gatsby-theme-newrelic/src/components/SEO.js
roadlittledawn: i ran the the build for the demo workspace and didn't see any errors. think that's sufficient?

jerelmiller: In this case, I'm not worried about errors, I'm worried that the value will be either `null` or `undefined` (depending on how that location comes back in the SSR case). Since `location` is an object, accessing the `hostname` property is always going to succeed.

packages/gatsby-theme-newrelic/src/components/SEO.js
jerelmiller: I'd opt for removing the meta tag. If we want to "auto detect" the type rather than make it configurable, I think it makes total sense that we'd opt for not having it, rather than being an incorrect value.

packages/gatsby-theme-newrelic/src/components/SEO.js
roadlittledawn: ah yes. thanks for catching that, @jerelmiller ! `location.hostname` is indeed `undefined` during SSR. i know we've done similar things to handle something like this during SSR. i'll look for that. in the meantime, let me know how you've handled that in the past / know where i should look

roadlittledawn: i suppose i could use the `siteUrl` from the static query in the component. unless there's something more reliable

roadlittledawn: i went that route of using `siteUrl` since that should be fairly reliable AND i overlooked it being used for other `<link>` elements in the component

LizBaker (approved):

PR #324, opened 2021-03-12
chore: adds issuelabel prop to global footer

jpvajda:
This PR adds an issueLabels prop to the global footer so we can define the type of label that is attached to an issue when a user clicks create issue on the Docs Site. Given the changes implemented here:

https://github.com/newrelic/docs-website/issues/1206

I think this will just `work` after this is merged and we bump the theme on the docs site. 🤞

### DIFF
--- a/packages/gatsby-theme-newrelic/src/components/GlobalFooter.js
+++ b/packages/gatsby-theme-newrelic/src/components/GlobalFooter.js
@@ -16,7 +16,12 @@ import Link from './Link';
 // the footer
 const copyrightSymbol = String.fromCharCode(169);
 
-const GlobalFooter = ({ fileRelativePath, className, pageTitle }) => {
+const GlobalFooter = ({
+  fileRelativePath,
+  className,
+  pageTitle,
+  issueLabels,
+}) => {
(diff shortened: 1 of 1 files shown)


packages/gatsby-theme-newrelic/src/components/GlobalFooter.js
```diff
@@ -99,6 +104,7 @@ const GlobalFooter = ({ fileRelativePath, className, pageTitle }) => {
               fileRelativePath={fileRelativePath}
               variant={Button.VARIANT.OUTLINE}
               size={Button.SIZE.SMALL}
+              labels={issueLabels}
```
LizBaker, line 107: We'll need the labels prop to be in the `CreateIssueButton` component above, not this `EditPageButton`

LizBaker (changes requested):
Left a comment to switch which component is consuming the new prop, and also we'll wanna add issueLabels to the proptypes on lines 201-205 like so `issueLabels: CreateIssueButton.propTypes.labels`

LizBaker (approved):
✨

PR #327, opened 2021-03-15
fix: Add optional issueLabels prop to global footer

jpvajda:
This PR correctly submits the changes from this original PR #324 with fix prefix in order to trigger a version bump on the theme.


LizBaker (approved):
💯

jerelmiller:
I wonder if we should use a tool like [commitlint](https://github.com/conventional-changelog/commitlint) to catch this earlier on since the theme requires conventional commits for versioning.

jpvajda:
@jerelmiller yes please. DevToolKit has super strict linting rules on commit messages. at first I hated it, then I loved it. 🤣

PR #341, opened 2021-04-11
fix: removes utm source from theme

jpvajda:
This PR removes the use of UTM source from the theme as that is causing trouble in the analytics pipeline for tracking user clicks on the sign up buttons on the Docs, Dev, and OSS sites.  This also includes formatting changes to the readme file from markdown linting.


packages/gatsby-theme-newrelic/README.md
```diff
@@ -9,100 +9,101 @@ websites](https://opensource.newrelic.com).
...
-  - [`useUserId`](#useuserid)
-  - [`usePrevious`](#useprevious)
-- [I18n](#i18n-1)
-- [Announcements](#announcements)
-- [Utils](#utils)
-  - [`formatCode`](#formatcode)
-- [Testing](#testing)
-  - [Changes to Jest config](#changes-to-jest-config)
-  - [Mocking `ReactDOM.createPortal` for snapshot testing](#mocking-reactdomcreateportal-for-snapshot-testing)
+- [`gatsby-theme-newrelic`](#gatsby-theme-newrelic)
```
zstix, line 12: Do we need to indent the _entire list_ for a single bullet point? The more indentation we add to this list, the more error prone it becomes ("am I missing a space?")

zstix (approved):
One nitpick comment about the README, but the code looks good ✅

packages/gatsby-theme-newrelic/README.md
jpvajda: I use a Markdown linter and  a MD table generator, so it's all in the best practice of markdown. It auto corrects on save of a file, but I can modify this if you feel it's an issue.

zstix: I don't feel strongly about it, we can leave it in.

That said, this isn't best practice because there isn't a best practice with markdown lists. Your linter formats them this way because your plugin's author has an opinion on how this should look. The rest of the team _doesn't_ have that plugin, which is why it _isn't_ currently indented.

PR #343, opened 2021-04-12
Add google tracking manually if not using tessen

roadlittledawn:
## Description

This manually adds google tag manager tracking and, if a user opts out / has opted out of cookies (and thus Tessen is not used at all), it calls `gtag()` to send page views.

This also adds a NR Browser `PageAction` and a GA event to report when users click `No`.

For good measure, I also updated Segment sources for each of the sites to anonymize IP addresses in GA (if brought in by Tessen) and exclude IPs from Snowflake.

## Sample site config
```
googleTagManager: {
  trackingId: 'ID_HERE',
  src: 'https://www.googletagmanager.com/gtag/js',
  options: {
    anonymize_ip: true,
  },
},
```

## Related issues
* Closes https://github.com/newrelic/docs-website/issues/1331

### DIFF
--- a/packages/gatsby-theme-newrelic/gatsby-node.js
+++ b/packages/gatsby-theme-newrelic/gatsby-node.js
@@ -9,6 +9,7 @@ const {
   getTrailingSlashesConfig,
   getResolvedEnv,
   getI18nConfig,
+  getGtmConfig,
 } = require('./src/utils/config');
 const pageTransforms = require('./gatsby/page-transforms');
 const { TESSEN_PATH } = require('./gatsby/constants');
@@ -126,6 +127,7 @@ exports.sourceNodes = (
   const { relatedResources } = withDefaults(themeOptions);
   const { createNode } = actions;
   const tessen = getTessenConfig(themeOptions);
+  const googleTagManager = getGtmConfig(themeOptions);
   const env = getResolvedEnv(themeOptions);
   const { forceTrailingSlashes } = getTrailingSlashesConfig(themeOptions);
 
@@ -155,6 +157,13 @@ exports.sourceNodes = (
     },
     tessen: tessen
       ? { product: tessen.product, subproduct: tessen.subproduct }
+      : null,
+    googleTagManager: googleTagManager
+      ? {
+          trackingId: googleTagManager.trackingId,
+          src: googleTagManager.src,
+          options: googleTagManager.options,
+        }
       : null,
   };
 
--- a/packages/gatsby-theme-newrelic/gatsby-node.js
+++ b/packages/gatsby-theme-newrelic/gatsby-node.js
@@ -158,13 +158,7 @@ exports.sourceNodes = (
     tessen: tessen
       ? { product: tessen.product, subproduct: tessen.subproduct }
       : null,
-    googleTagManager: googleTagManager
-      ? {
-          trackingId: googleTagManager.trackingId,
-          src: googleTagManager.src,
-          options: googleTagManager.options,
-        }
-      : null,
+    googleTagManager: googleTagManager || null,
   };
 
   createNode({
--- a/packages/gatsby-theme-newrelic/gatsby/on-pre-render-html.js
+++ b/packages/gatsby-theme-newrelic/gatsby/on-pre-render-html.js
@@ -7,20 +7,18 @@ const onPreRenderHTML = (
   { getHeadComponents, replaceHeadComponents },
   themeOptions
 ) => {
-  const googleTagManager = getGtmConfig(themeOptions);
   let gtagScript;
-  let googleTrackScript;
+  const googleTagManager = getGtmConfig(themeOptions);
 
-  if (themeOptions.googleTagManager) {
-    gtagScript = (
-      <script
-        async
-        key="nr-gtag"
-        src={`${googleTagManager.src}?id=${googleTagManager.trackingId}`}
-      />
-    );
+  gtagScript = googleTagManager ? (
+    <script
+      async
+      key="nr-gtag"
+      src={`${googleTagManager.src}?id=${googleTagManager.trackingId}`}
+    />
+  ) : null;
 
-    const scriptStr = `
+  const scriptStr = `
     var options = {
       send_page_view: false,
       anonymize_ip: true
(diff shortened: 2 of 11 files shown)


packages/gatsby-theme-newrelic/gatsby-node.js
```diff
@@ -156,6 +158,13 @@ exports.sourceNodes = (
     tessen: tessen
       ? { product: tessen.product, subproduct: tessen.subproduct }
       : null,
+    googleTagManager: googleTagManager
+      ? {
+          trackingId: googleTagManager.trackingId,
+          src: googleTagManager.src,
+          options: googleTagManager.options,
+        }
+      : null,
```
zstix, line 167: I _think_ we could simplify this to:
```suggestion
    googleTagManager: googleTagManager || null,
```

packages/gatsby-theme-newrelic/src/utils/page-tracking/gtag.js
```diff
@@ -0,0 +1,69 @@
...
+    "You have enabled Google Tag Manager, but it is misconfigured. This route change has not been tracked. Please configure the 'googleTagManger.src' and 'googleTagManger.trackingId' options in gatsby-config.js"
+  );
+};
+
+const canSendPageView = (config) => config && config.src && config.trackingId;
+
+const sendPageview = ({ location }, themeOptions) => {
+  // wrap inside a timeout to make sure react-helmet is done with its changes (https://github.com/gatsbyjs/gatsby/issues/11592)
+  requestAnimationFrame(() => {
+    requestAnimationFrame(() => {
```
zstix, line 23: Out of curiosity, why did we go with nested `requestAnimationFrame` calls over [`useEffect` and `setTimeout`](https://github.com/nfl/react-helmet/issues/189#issuecomment-698953347)?

zstix (commented):
Couple of minor questions, but this looks great!!!

packages/gatsby-theme-newrelic/src/utils/page-tracking/gtag.js
roadlittledawn: i wondered the same, but i was just following suit from what we used for Tessen calls in https://github.com/newrelic/gatsby-theme-newrelic/blob/72f6c726888f494ebedf57196ce38ac63a630a4b/packages/gatsby-theme-newrelic/src/utils/page-tracking/tessen.js#L43

zstix (approved):

packages/gatsby-theme-newrelic/src/utils/page-tracking/gtag.js
jerelmiller: I might be able to answer this one for you:

We can't use `useEffect` here because we aren't executing this code inside a React component. Hooks only work inside React components. the `onRouteUpdate` Gatsby hooks is just a normal function.

As per the `requestAnimationFrame`, I noticed this in the [`gatsby-plugin-google-gtag`](https://github.com/gatsbyjs/gatsby/blob/54d4721462b9303fed723fdcb15ac5d72e103778/packages/gatsby-plugin-google-gtag/src/gatsby-browser.js#L21-L28) plugin and figured they knew something I didn't 😄.

packages/gatsby-theme-newrelic/gatsby/type-defs.graphql
```diff
@@ -52,3 +53,13 @@ type NewRelicThemeTessenConfig {
...
   subproduct: String
 }
+
+type googleTagManagerConfig {
+  trackingId: String
+  src: String
+  options: googleTagManagerOptions
+}
+
+type googleTagManagerOptions {
```
jerelmiller, line 63:
```suggestion
type GoogleTagManagerConfig {
  trackingId: String
  src: String
  options: GoogleTagManagerOptions
}

type GoogleTagManagerOptions {
```

GraphQL types are standardized around capital names to denote them as types rather than variables.

packages/gatsby-theme-newrelic/gatsby/type-defs.graphql
```diff
@@ -52,3 +53,13 @@ type NewRelicThemeTessenConfig {
...
+  anonymize_ip: Boolean
```
jerelmiller, line 64: Out of curiosity, why is this snake_case vs camelCase?

packages/gatsby-theme-newrelic/src/utils/config/gtag.js
```diff
@@ -0,0 +1,9 @@
+const { buildConfigGetter } = require('../configBuilder');
+
+module.exports = buildConfigGetter('googleTagManager', {
+  envOptions: false,
+  defaults: {
+    minify: process.env.NODE_ENV !== 'development',
```
jerelmiller, line 6: It doesn't look like you use this, so you can omit this.

packages/gatsby-theme-newrelic/gatsby/on-pre-render-html.js
```diff
@@ -1,11 +1,42 @@
...
 const onPreRenderHTML = (
   { getHeadComponents, replaceHeadComponents },
   themeOptions
 ) => {
+  const googleTagManager = getGtmConfig(themeOptions);
+  let gtagScript;
+  let googleTrackScript;
+
+  if (themeOptions.googleTagManager) {
+    gtagScript = (
```
jerelmiller, line 15: [nit] It doesn't look like these really need to be in a conditional since you're checking them anyways [on line 61 and 62](https://github.com/newrelic/gatsby-theme-newrelic/pull/343/files#diff-2ce4e67e6fc3cf9f2b0afa489dce365fa84ce3014b74049c9f9e5eda2556f4daR61). I'd just pull these out to a `const` and remove this conditonal.

jerelmiller:
Added a few suggestions for you 😄 . Feel free to move forward without my approval if so you don't get help up. Figured I'd at least comment on a couple places where you might be able to provide a bit of improvement.

packages/gatsby-theme-newrelic/gatsby/type-defs.graphql
roadlittledawn: oh i just mirrored the syntax expected in the `gtag()` options arg.

packages/gatsby-theme-newrelic/gatsby/type-defs.graphql
roadlittledawn: not sure how i missed that 🤦  thanks for catching that!

roadlittledawn:
> Added a few suggestions for you 😄 . Feel free to move forward without my approval if so you don't get help up. Figured I'd at least comment on a couple places where you might be able to provide a bit of improvement.

Thanks a heap, @jerelmiller !  Much appreciated :)

zstix (approved):

PR #368, opened 2021-04-22
Fix referral field not being sent to analytics

aswanson-nr:
# Summary
The `referrer` field was not being sent when a user clicks the `free account` button on our sites. This was caused by the `rel="noreferrer"` property on our the SignUpLink component.
The `noreferrer` and `noopener` types control the information passed to the new tab. The former prevents referrer information from being passed to the new tab and the latter prevents a lot of other information from being passed.

We have 2 main options to fix this:
1. Remove the `noreferrer` type (possible security risk, but we do own the site we are linking to)
2. Do not open the link in a new tab

This PR currently fixes this issue through option 1.


aswanson-nr:
The issue https://github.com/newrelic/docs-website/issues/1816

nr-kkenney (approved):
lgtm

PR #375, opened 2021-04-30
New search experience 

caylahamann:
## Summary 
Brand new search experience hot off the presses! This brings: 

* Full component control over search (no more SwiftypeSearch)
* A BRAND NEW look and experience, much more in line with modern experiences like Algolia and Stripe
* Nicer mobile styles for results

## Screenshots

Screen Shot 2021-04-30 at 3 59 43 PM
Screen Shot 2021-04-30 at 4 00 57 PM

Screen Shot 2021-04-30 at 3 59 50 PM
Screen Shot 2021-04-30 at 4 00 03 PM
Screen Shot 2021-04-30 at 4 00 15 PM


roadlittledawn:
say whaaa?! awesome! 👏

austin-schaefer:
OMG very excited about this! Would love to test it out.

austin-schaefer:
Hi @caylahamann / @jerelmiller , quick question:

> Full component control over search (no more SwiftypeSearch)

We're still using the underlying swiftype endpoints (and logic) here, right? Just changing how we display results on our end?

jerelmiller:
@austin-schaefer 

> We're still using the underlying swiftype endpoints (and logic) here, right? Just changing how we display results on our end?

Thats (mostly) correct! Still using Swiftype, but we stopped using the elastic React components that were super hard for us to work with (no more using [this package](https://github.com/elastic/search-ui)). We now do all of the fetching ourselves to Swiftype, using the same endpoints as before, just with our own components.

When you say "logic", what are you referring to here? I can say we only fetch results for the docs, dev, and open source sites and nothing else. Is this what you're referring to? This is the only reason I say "mostly" correct since I'm not totally sure what you mean here.

jpvajda:
I'm so excited for this new search expeirence! cc @polfliet / @austin-schaefer 🚀

roadlittledawn:
@caylahamann || @jerelmiller . looking perdy nifty so far! one thing i've noticed thus far is if you request a page with a search set in the query params like http://localhost:8001/?q=alerts , page looks like this. this happened while running dev server locally.

Screen Shot 2021-05-04 at 15 10 22

jerelmiller:
How strange! We used that query param during the implementation since we could make a code change, have the page reload, and see the modal. We can figure out what we broke in one of our recent commits. Thanks for checking this out!

roadlittledawn:
@jerelmiller  not sure if this is at play, but we did change the import in `useQueryParams.js` from `useNavigate` to `navigate` during the Gatsby v3 upgrade. the `useNavigate` hook was causing errors, but `navigate` didn't so we went that

jerelmiller:
Good to know! I bet that has something to do with it. The last commit was the merge that included all the Gatsby 3 stuff, so that would make a lot of sense. We will take a look a bit further!

packages/gatsby-theme-newrelic/src/components/Link.js
```diff
@@ -13,98 +13,86 @@ const isExternal = (to) => to.startsWith('http');
 const isNewRelic = (to) => to.startsWith('https://newrelic.com');
 const isSignup = (to) => to.startsWith('https://newrelic.com/signup');
 
-// https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a
-const ATTRIBUTES = [
```
jerelmiller, line 17: This is missing all of the event handlers for the `Link`, which caused the hover styling to break on the new search experience. Because there are lots of different handlers that can be used for links, I'd rather find and fix the areas of the apps that send invalid props to the `Link` component. Thats going to be easier than trying to maintain a whitelist of "valid" props here, since we don't have to track all of the different handlers that can be passed to these components (e.g. `onMouseOver`, `onMouseOut`, etc.).

jpvajda:
It looks like we have one final issue to work out that @jerelmiller 👋 is looking into.

jpvajda:
I followed up with @jerelmiller to see if he has time to finish this up, or if he preferred we did the final changes.

jpvajda:
From @jerelmiller 

> I think the only thing left is the language specific search. It’s just a matter of getting the locale and passing the right filters on search.

jerelmiller:
@jpvajda @roadlittledawn should be all fixed now! Locale is now taken into place. Thanks for catching this!

roadlittledawn (approved):
taking into account locale to filter on `type` look good to me! thanks a heap, @caylahamann && @jerelmiller !   :shipit:

PR #378, opened 2021-05-03
Add `customer_user_id` in all tessen calls

roadlittledawn:
## Description

Gets the value of `ajs_user_id` if available and sends along with all `tessen.page()` and `tessen.track()` calls. If no value available, sends empty string.

Note: we only use tessen when a user has agreed to cookie usage.

## Review notes

Testing is a little tricky since the `ajs_user_id` cookie is set to newrelic domains. Need to verify once one of the sites incorporates it. I tested locally with `ajs_anonymous_id` and it did work.

## Related issues
Closes https://github.com/newrelic/docs-website/issues/1668


nr-kkenney (approved):
lgtm

PR #386, opened 2021-05-06
fix: add 'env' to tessen call

nr-kkenney:


aswanson-nr (approved):
LGTM

packages/gatsby-theme-newrelic/src/utils/page-tracking/tessen.js
```diff
@@ -50,12 +50,7 @@ const trackViaTessen = ({ location }, themeOptions) => {
 
 const trackPageView = ({ config, env, location }) => {
   const { pageView } = config;
-  const {
-    name,
-    category,
-    getProperties = () => ({}),
```
jerelmiller, line 56: Out of curiosity, is `getProperties` used to set anything else besides `location` and `env`? I initially authored it this way to allow you to add additional custom properties in the tessen call, given a particular `location` and `env` (it didn't have to be just those 2 properties).

Totally fine if there is no use for this, but just wanted to point out that the intention here was to allow the possibility to customize additional properties when needed.

nr-kkenney: it is set up identically across the 3 places we use it:
```
    env: env === 'production' ? 'prod' : env,
    location: location.pathname
```
no extra values or customization beyond env and location 🙈 .

zstix (approved):

PR #387, opened 2021-05-06
Fixing console errors

zstix:
## Description
A lot of console errors & warnings come up while browsing the docs site. It turns out that many of these are coming from the theme itself. This PR aims to address some of these issues. You should be able to check out this branch, run `yarn start` and not run into any _errors_. There are a couple warnings that I've been unable to remedy:

* For some reason, eslint is not respecting our rules about `react/jsx-pascal-case`
* There is some random file that's importing a variable with the name `version`. I've been unable to find it.

In addition to those two warnings, there's also an occasional bug in the console about a `<div>` that doesn't have a matching `</div>`. I'm still working to track that down, but I haven't been able to find the exact location yet.

## Related Issue(s)
* Relates to newrelic/docs-website#997

## Screenshot(s)
Here's a few of the errors that we should not longer be seeing:
Screen Shot 2021-05-06 at 1 49 14 PM
Screen Shot 2021-05-06 at 4 01 05 PM
Screen Shot 2021-05-06 at 4 01 19 PM
Screen Shot 2021-05-06 at 4 04 04 PM

### DIFF
--- a/demo/src/layouts/MainLayout.js
+++ b/demo/src/layouts/MainLayout.js
@@ -24,8 +24,8 @@ const MainLayout = (props) => {
       <SEO location={location} />
       <GlobalHeader />
       <MobileHeader>
-        {nav.map((page) => (
-          <NavItem key={page.url} page={page} />
+        {nav.map((page, index) => (
+          <NavItem key={`mobile${index}${page}`} page={page} />
         ))}
       </MobileHeader>
       <Layout>
@@ -44,8 +44,8 @@ const MainLayout = (props) => {
             `}
           />
           <Navigation searchTerm={searchTerm}>
-            {nav.map((page) => (
-              <NavItem key={page.url} page={page} />
+            {nav.map((page, index) => (
+              <NavItem key={`nav${index}${page}`} page={page} />
             ))}
           </Navigation>
         </Layout.Sidebar>
--- a/packages/gatsby-theme-newrelic/src/components/MarkdownContainer.js
+++ b/packages/gatsby-theme-newrelic/src/components/MarkdownContainer.js
@@ -15,7 +15,7 @@ const MarkdownContainer = ({
         --block-element-spacing: 1.5rem;
         --text-spacing: var(--paragraph-spacing);
 
-        > *:first-child {
+        > *:first-of-type {
           margin-top: 0;
         }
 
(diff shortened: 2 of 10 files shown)


packages/gatsby-theme-newrelic/src/components/MarkdownContainer.js
```diff
@@ -39,13 +39,13 @@ const MarkdownContainer = ({
           margin-bottom: var(--text-spacing);
         }
 
-        h2:not(:first-child) {
+        h2:not(:first-of-type) {
```
jerelmiller, line 42: I know the warnings are annoying, but in general, be careful with these changes. We want to make sure to test them thoroughly, otherwise you can end up with weird bugs. `first-child` and `first-of-type` are not the same thing. Take for example the following:

```js
<section className="container">
  <div className="divider" />
  <h2>First title</h2>
  <p>Stuff</p>
  <h2>Second title</div>
</div>
```

And the following CSS:

```js
h2 {
  margin-top: 1rem;
}

.container h2:first-child {
  margin-top: 0;
}
```

Given this CSS, we'd expect that the "First title" `h2` would have `1rem` of margin above it, because we've said that we _only_ want to remove it if its the first child in the container.

Now lets change the CSS to this:

```css
h2 {
  margin-top: 1rem;
}

.container h2:first-of-type {
  margin-top: 0;
}
```

Rather than removing the space only when the `h2` is the first child, you are removing the spacing from the first _instance_ of it. If the intention is to keep the space above the `h2` when its not the first child, then this updated CSS has broken that.

I would definitely vet these changes pretty thoroughly as it could collapse/add space where you don't expect.

zstix: Ah, thanks for catching that. I thought I tested it thoroughly, but it would appear like I've missed some cases.

I might just back this out for now since the real reason I made this PR was to fix the errors, not the warnings.

nr-kkenney (dismissed):
i feel good approving after zack's walkthrough.

demo/src/layouts/MainLayout.js
```diff
@@ -24,8 +24,8 @@ const MainLayout = (props) => {
       <SEO location={location} />
       <GlobalHeader />
       <MobileHeader>
-        {nav.map((page) => (
-          <NavItem key={page.url} page={page} />
+        {nav.map((page, index) => (
+          <NavItem key={`mobile${index}${page}`} page={page} />
```
LizBaker, line 28: does it make sense for this to also use page.url (`mobile${index}${page.url}`)? instead of the whole page object

zstix: I think we really only need the index, but I can add that url to make it more specific.

LizBaker (approved):

PR #395, opened 2021-05-12
Fixing search result links

zstix:
## Description
A fix that allows us to set the innerHTML for links. This is necessary to resolve an issue with the search results not showing link text.


nr-kkenney (approved):
works on my machine

PR #398, opened 2021-05-14
Remove split button spinner

LizBaker:
Removing the spinner for split button color change


aswanson-nr:
This looks great to me! I just have two things on my mind
1. I vaguely remember talking about using an outline button for the initial state. @zstix Do you remember talking about that?
2. We probably shouldn't release this until the current testing period is over, since we don't want to affect the results. Or just release and restart the test.

aswanson-nr (approved):

PR #400, opened 2021-05-20
Fix: Bump SearchModal z-index up higher

roadlittledawn:
## Description

On PR review for opensource, there are some elements that have a 300 z-index, causing to float over the search modal. This bumps z-index up much higher to avoid that since it needs to float over most things.

## Screenshot

Screen Shot 2021-05-19 at 11 34 17 AM

## Related PR
https://github.com/newrelic/opensource-website/pull/784


nr-kkenney (approved):
looks good

PR #401, opened 2021-05-20
chore: Adding demo examples to Readme

rudouglas:
Including images


README.md
```diff
@@ -15,24 +20,90 @@ yarn
...
+
+### Example
+
+For this example we will show how to add some content to a page and see the render results in the browser. We will use the `build-hello-world-app` document for testing, so with the **develop** server still running, navigate to and open the following in your favourite code editor:
+
+ - `demo/src/content/build-apps/build-hello-world-app.mdx`
+
+You should see something like the following (possibly less wine coloured):
+
+<img src="packages/gatsby-theme-newrelic/src/images/example-mdx.png" alt="drawing" width="75%"/>
```
jerelmiller, line 57: Since these images are for the README, could you move them to a top-level folder in this repo instead of the theme package itself? The `src` directory is packaged with the theme in NPM, so we want to prevent those images from being downloaded as well (which is just wasted space).

rudouglas: Good to know, of course will do :D

rudouglas: done @jerelmiller :D

jerelmiller: Thank you!!

aswanson-nr (approved):

PR #402, opened 2021-05-20
Fix to open image links in new tab

roadlittledawn:
## Description

Adds a check for the class that the `gatsby-remark-images` plugin adds to `<a>` tags that wrap images and returns straight up `<a>` instead of a Gatsby `<Link />` component.

Since we're explicitly using `className` now, had to set that explicitly in all returns otherwise it does not get passed via `...props`

## Screenshots

https://user-images.githubusercontent.com/2952843/119009206-bd9e0b00-b947-11eb-8457-fb0f45d1e2c8.mp4

## Related issues

https://github.com/newrelic/docs-website/issues/2363

### DIFF
--- a/packages/gatsby-theme-newrelic/src/components/Link.js
+++ b/packages/gatsby-theme-newrelic/src/components/Link.js
@@ -14,7 +14,7 @@ const isNewRelic = (to) => to.startsWith('https://newrelic.com');
 const isSignup = (to) => to.startsWith('https://newrelic.com/signup');
 
 const Link = forwardRef(
-  ({ to, onClick, instrumentation = {}, ...props }, ref) => {
+  ({ to, onClick, instrumentation = {}, className, ...props }, ref) => {
     const locale = useLocale();
 
     const {
@@ -49,7 +49,7 @@ const Link = forwardRef(
     }
 
     if (isHash(to)) {
-      return <a ref={ref} href={to} {...props} />;
+      return <a ref={ref} href={to} className={className} {...props} />;
     }
 
     if (isSignup(to)) {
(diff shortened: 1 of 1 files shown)


packages/gatsby-theme-newrelic/src/components/Link.js
```diff
@@ -81,13 +83,18 @@ const Link = forwardRef(
       );
     }
 
+    if (className === 'gatsby-resp-image-link') {
```
jerelmiller, line 86: [nit] If you didn't want to destructure + add the `className` to everything, you could also do this trick:

```js
if (props.className === 'gatsby-resp-image-link')
```

then leave everything else as it was.

jerelmiller: You might also consider wrapping this in a helper function to make it clear what you're wanting to do here.

```js
// or whatever makes sense to pass in here
if (isImageLink({ ...props, to }))
```

roadlittledawn: i initially used `props.className` but got prop type validation linting error 🤷 

good call on moving to function like the others

jerelmiller: I think if you add the `propType` itself, that should resolve that eslint warning without needing to destructure the prop in the component. I forgot about that rule 😆

roadlittledawn: oh didn't realize i could just add `className` to `Link.propTypes`. it passed the linting check and classes are applied to other links as expected (tested change in local docs repo as well).

jerelmiller (approved):
Nice! Thanks for making those updates! This simplifies things a bit more :)

PR #407, opened 2021-05-21
fix: handle lack of access to local storage

aswanson-nr:
# Summary
* Remove direct interaction with local storage in useUserId and use
  `use-local-storage-state`
* Replace use of `use-persisted-state` in the AnnouncementBanner with
  `use-local-storage-state`
* `on-client-entry.js` add check for local storage before accessing it
* Add check for local storage and in-memory store to DarkModeToggle
* Uses of `use-persisted-state` were removed because the library does
  not have a fallback if local storage is unavailable

### DIFF
--- a/packages/gatsby-theme-newrelic/src/utils/isLocalStorageAvailable.js
+++ b/packages/gatsby-theme-newrelic/src/utils/isLocalStorageAvailable.js
@@ -1,0 +1,12 @@
+const isLocalStorageAvailable = () => {
+  try {
+    localStorage.setItem('test', 'value');
+    localStorage.getItem('test');
+    localStorage.removeItem('test');
+  } catch (error) {
+    return false;
+  }
+  return true;
+};
+
+export default isLocalStorageAvailable;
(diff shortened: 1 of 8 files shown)


aswanson-nr:
Related newrelic/docs-website#2293

aswanson-nr:
@jerelmiller I would love to get your thoughts on this

jerelmiller:
Sure thing! I'll take a look at this when I get a chance

packages/gatsby-theme-newrelic/src/utils/isLocalStorageAvailable.js
```diff
@@ -0,0 +1,12 @@
+const isLocalStorageAvailable = () => {
```
zstix, line 1: I'm not 100% sure, but this might be simpler as `'setItem' in localStorage && 'getItem' in localStorage`

aswanson-nr: I like it! Although it looks like the try/catch are still required because certain browsers (looking at you, Brave) throw an error instead of just removing localStorage from the global scope.

LizBaker (approved):

PR #409, opened 2021-05-26
Improve search results by removing field weights

roadlittledawn:
## Summary

Removes the weighting fields and lets the swiftype admin UI setting apply instead.

## Related issues
Closes https://github.com/newrelic/docs-website/issues/2455

### DIFF
--- a/packages/gatsby-theme-newrelic/src/components/SearchModal/search.js
+++ b/packages/gatsby-theme-newrelic/src/components/SearchModal/search.js
@@ -24,9 +24,6 @@ const search = async ({ searchTerm, filters = {}, perPage = 10, page = 1 }) => {
           },
         },
       },
-      search_fields: {
-        page: ['title', 'body'],
-      },
       filters: {
         ...filters,
         page: {
--- a/packages/gatsby-theme-newrelic/src/components/SearchModal/search.js
+++ b/packages/gatsby-theme-newrelic/src/components/SearchModal/search.js
@@ -25,7 +25,7 @@ const search = async ({ searchTerm, filters = {}, perPage = 10, page = 1 }) => {
         },
       },
       search_fields: {
-        page: ['title^2', '*'],
+        page: ['title', 'body'],
       },
       filters: {
         ...filters,


austin-schaefer:
Hi @roadlittledawn , added a couple of test queries that might be useful to look at in #2455.

jerelmiller:
@roadlittledawn would it be better to just remove that `search_fields` field altogether? Unless we want to restrict the search to only look at `title` and `body` (which is also ok)

roadlittledawn:
@jerelmiller was wondering that myself. the swiftype UI has couple more field weight settings for:
- `sections`
- `updated_at` (determined by crawler since we don't explicitly set it)
- `published_at` (determined by crawler since we don't explicitly set it)

not sure what impact that would have on results, but i know `title` and `body` are probably most relevant

roadlittledawn:
> Hi @roadlittledawn , added a couple of test queries that might be useful to look at in #2455.

@austin-schaefer here's those results

roadlittledawn:
removing the `search_fields` altogether seems to get better matches for `csv`. `desc` doesn't seem to be any better

roadlittledawn:
just pushed commit that removes the specific `search_fields` altogether because results seemed as good better, the little bit i did.

@mmfred && @austin-schaefer let me know what you think of results now

mmfred:
> @mmfred && @austin-schaefer let me know what you think of results now

Definitely seems to be improved. I didn't find results that locate "desc" or "csv" which someone reported as missing from search this morning. However, was able to find other strings that have been requested. And a few others.

roadlittledawn:
> > @mmfred && @austin-schaefer let me know what you think of results now
> 
> Definitely seems to be improved. I didn't find results that locate "desc" or "csv" which someone reported as missing from search this morning. However, was able to find other strings that have been requested. And a few others.

@mmfred do we know what doc they were looking for when searching `csv`? do any of the results in the screenshot above look right?

mmfred:
@clinton - I don't know what they were looking for specifically, but yes, that screenshot seems like it would do the trick.

jerelmiller:
> the swiftype UI has couple more field weight settings

This is probably the better approach anyways than doing code-level weighting. That way anybody with access to the Swiftype console can update the weights at any given time. Thanks for removing that and getting to this so quickly!

LizBaker (approved):

austin-schaefer:
Thanks @roadlittledawn !

austin-schaefer:
FWIW I think the CSV results are because we don't currently have a doc that clearly highlights CSV export from dashboard. So I'm realizing that one is a red herring.

PR #410, opened 2021-05-27
fix: remove mocked storage and use built in option

aswanson-nr:
# Summary
Remove mocked localStorage since it causes issues on the Docs site. Passing in `false` causes `useDarkMode` to default to a React hook instead.


aswanson-nr:
More work for https://github.com/newrelic/docs-website/issues/2293

roadlittledawn (approved):
didn't realize that `storageProvider` option existed for `use-dark-mode`. 

collapsers worked and so did search with cookies disabled. didn't notice any other anomalies

rudouglas:
I spun this up and all seemed to work correctly 🍾

PR #417, opened 2021-06-11
fix: add pageaction to track user search terms

nr-kkenney:
* added a pageaction to track user terms in the globalheader since that
has a debounce hook. this should enable search terms data populating our
dashboard again.

Relates to: https://github.com/newrelic/docs-website/issues/2598


rudouglas (approved):
Looks G 2 Me

PR #424, opened 2021-07-13
Explicilty set referrer URL path for Tessen

roadlittledawn:
## Description

Tessen is seemingly not capturing most referrer URL paths if a user has loaded the site and then clicks on any links for other pages on the site. We can explicitly set that, because we already send a "pageview" event for each route change. This just adds a `referrer` value to the `properties` in the tessen payload.

Also includes remove superfluous `location.pathname` since that's captured via `gatsby-config.js` properties function. Also, it seems `location` is used by tessen to indicate it's a `Public` site, so it was being overridden anyway.

## Related issues / PRs
Related to https://github.com/newrelic/docs-website/issues/2841


aswanson-nr (commented):
LGTM!

jpvajda (approved):

PR #429, opened 2021-07-20
Generate correct path on Windows

vladar:
`path.join` is platform-dependent: it uses backslashes as separators on Windows.

So pages generated with this theme have incorrect URLs.


jpvajda:
Thanks for the contribution, we'll take a look at this!

zstix (approved):

PR #449, opened 2021-09-03
chore: add io to global header nav

rudouglas:
Because the nav links to https://observabilitypacks.gatsbyjs.io/instant-observability/ it won't be possible to see the darker background until the pr is merged
Screenshot 2021-09-03 at 13 56 28

### DIFF
--- a/packages/gatsby-theme-newrelic/src/components/GlobalHeader.js
+++ b/packages/gatsby-theme-newrelic/src/components/GlobalHeader.js
@@ -232,7 +232,7 @@ const GlobalHeader = ({ className }) => {
                 <Dropdown.MenuItem href="https://learn.newrelic.com/">
                   Learn
                 </Dropdown.MenuItem>
-                <Dropdown.MenuItem href="https://observabilitypacks.gatsbyjs.io/instant-observability/">
+                <Dropdown.MenuItem href="https://newrelic.com/instant-observability/">
                   Instant Observability
                 </Dropdown.MenuItem>
               </Dropdown.Menu>
@@ -287,7 +287,7 @@ const GlobalHeader = ({ className }) => {
                 </GlobalNavLink>
               </li>
               <li>
-                <GlobalNavLink href="https://observabilitypacks.gatsbyjs.io/instant-observability/">
+                <GlobalNavLink href="https://newrelic.com/instant-observability/">
                   Instant Observability
                 </GlobalNavLink>
               </li>
--- a/packages/gatsby-theme-newrelic/src/components/GlobalHeader.js
+++ b/packages/gatsby-theme-newrelic/src/components/GlobalHeader.js
@@ -232,6 +232,9 @@ const GlobalHeader = ({ className }) => {
                 <Dropdown.MenuItem href="https://learn.newrelic.com/">
                   Learn
                 </Dropdown.MenuItem>
+                <Dropdown.MenuItem href="https://observabilitypacks.gatsbyjs.io/instant-observability/">
+                  Instant Observability
+                </Dropdown.MenuItem>
               </Dropdown.Menu>
             </Dropdown>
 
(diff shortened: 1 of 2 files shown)


packages/gatsby-theme-newrelic/src/components/GlobalHeader.js
```diff
@@ -232,6 +232,9 @@ const GlobalHeader = ({ className }) => {
                 <Dropdown.MenuItem href="https://learn.newrelic.com/">
                   Learn
                 </Dropdown.MenuItem>
+                <Dropdown.MenuItem href="https://observabilitypacks.gatsbyjs.io/instant-observability/">
```
aswanson-nr, line 235: I think we may want to point this to the actual url, since it will be a hassle to get this updated again for the correct url.

aswanson-nr (commented):
Looks great! Just left a comment about the URL

packages/gatsby-theme-newrelic/src/components/GlobalNavLink.js
```diff
@@ -19,7 +20,12 @@ const GlobalNavLink = ({ children, href }) => {
     }
   `);
 
-  const isCurrentSite = href.startsWith(siteUrl);
+  const location = useLocation();
+  let isCurrentSite = href.startsWith(siteUrl);
+
+  if (location.pathname.includes('/instant-observability/')) {
+    isCurrentSite = href.includes('/instant-observability/');
+  }
```
zstix, line 28: If we're going to hard-code the path here (and I recommend we don't, see below), then we should opt for a ternary for `isCurrentSite` rather than a `let` we modify - it will help our code be more safe.

Furthermore, I'm not sure how I feel about hard-coding this logic. I think we might want to consider adopting the approach of [open for extension, but closed for modification](https://en.wikipedia.org/wiki/Open%E2%80%93closed_principle). If we modify this code and then we find we need to add _another_ site that needs to have a custom highlight, then we will need to modify this code again.

A way we could _extend_ this would be to add an optional prop to the parent component specifying which NavLink should be considered active:

```jsx
// where the component is called
<GlobalHeader active="https://newrelic.com/instant-observability" />

// in the parent component
<GlobalNavLink /* ..other props*/ active={active} />

// in this component
const isCurrentSite = href === active || (!href && href.startsWith(siteUrl);
```

Another way could be to handle it on the developer site itself using css. Something like this might work:

```css
a {
  /* reset styles */
}

a[href="https://newrelic.com/instant-observability"] {
  /* active state */
}
```

zstix (commented):
I've got a couple thoughts. I'll accept the suggestions right away, but I'd like to chat about how we extend / modify the `<GlobalHeader>` and `<GlobalNavLink>` components.

packages/gatsby-theme-newrelic/src/components/GlobalHeader.js
rudouglas: I chatted to John about this and the reason was to have a working link until we go live, i've also opened a catch-all ticket for the work to do this #450

packages/gatsby-theme-newrelic/src/components/GlobalNavLink.js
zstix: I put up [an alternative PR](https://github.com/newrelic/gatsby-theme-newrelic/pull/453) with these ideas. Let me know if you like one over the other.

rudouglas: Sorry meant to respond here yesterday, wanted to talk about this approach to see if it require changes to all sites

rudouglas: I do prefer the alternative approach and it sets us up better because we don't have to hardcode, I think Jerel is reviewing the other PR but i'll take a look as well. My preference would be the alternative anywho

rudouglas:
Closing this in favour of the refactor

Issue #438, opened 2021-08-24
[Public Catalog] Update global header for I/O


roadlittledawn:
## Summary

Update global header to include link to New Relic I/O in the nav menus (mobile and desktop).

### Mock Up 

<img width="215" alt="Screen Shot 2021-08-27 at 3 44 45 PM" src="https://user-images.githubusercontent.com/1395158/131191804-a998b659-4b9b-4be5-9dbe-2bc55fb9cbef.png">

## Acceptance criteria
- [x] When viewing either the quick starts landing or a details page, the New Relic I/O menu item has a darker background color (and not Developer)
- [x] The link element should be called `Instant Observability`
- [x] This element should be `LAST`  in the header

jpvajda:
Confirmed this should go last in the  nav order.

jpvajda:
To keep things simple and ensure that there are no issues with Google, we are going to leave I/O on developer.newrelic.com for the September launch. 

We'll use a wildcard redirect in tandem with the new marketplace site domain that is set to go live in October 
This will be a clean 301 redirect TO the developer.newrelic.com/instant-observability/ subfolder FROM newrelic.com/instant-observability) .

Therefore in the case of this button it should point to `https://developer.newrelic.com/instant-observability/` IMHO

rudouglas:
Cool thanks @jpvajda I might point it to `https://observabilitypacks.gatsbyjs.io/instant-observability/` for now so we have a working link and then add that as a change we need for production readiness after the  merge of the feature branch

jpvajda:
@rudouglas can u file a follow up ticket to set this correctly and label it `readiness`? - Thanks!

rudouglas:
Yessir
#450 
![](https://user-images.githubusercontent.com/17122543/132008657-408b7d09-6a2d-49dd-95ca-55082a4af8aa.png)

jpvajda:
Confirmed: The logo will be ready on Sept 10

PR #453, opened 2021-09-07
Add ability to custom-set active sites in `<GlobalHeader />`

zstix:
## Description
Adds some extra flexibility to our `<GlobalHeader />` component so that we can custom-set which page we would like to show as "active" (the one with a background color). We can use this code to update specific pages on the developer website to highlight "Instant Observability".

This PR also moves the storage of the links to an array so that, if we need to change the order / items, we only have to update it in one location (rather than a few places in the JSX).

## Reviewer Notes
This PR is an alternative approach compared to solving #438 (compared to #449). I don't mind which direction we go, just wanted to move the conversation forward.

### DIFF
--- a/packages/gatsby-theme-newrelic/README.md
+++ b/packages/gatsby-theme-newrelic/README.md
@@ -1270,63 +1270,20 @@ import { GlobalHeader } from '@newrelic/gatsby-theme-newrelic';
 
 **Props**
 | Prop | Type | Required | Default | Description |
-| --------- | ------ | -------- | ------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| `editUrl` | string | no | | Used by the edit page link in the global header to link to a GitHub URL where a user can edit the page's source. If omitted, the edit page link will be excluded. |
+| - | - | - | - | - |
+| `activeSite` | enum | no | | The New Relic site that we should present as "active". This should come from `NR_SITES`, which is exported. |
 
-**Gatsby configuration**
-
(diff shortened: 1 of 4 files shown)


jerelmiller:
I definitely prefer this approach over the other PR. I think it sets us up much better for long-term changes. If you're serious about going this route, I'd be happy to give this PR a full code review (I spotted a couple minor things in my quick glance at the code).

rudouglas:
Working on trying to refactor this today, there are a couple of issues that i've been able to fix, such as the `.map` on the object, i've changed that to a Set to maintain the order of the items.

The issue i'm stuck on is one I was seeing when trying not to hard-code a url, both the `Developer` and `Instant Observability` buttons are highlighted because they are on the same domain, taking another crack at it today
Screenshot 2021-09-09 at 13 50 10

rudouglas:
This approach works when you are on `Instant Observability` but when you are on `Developer` both are still highlighted and I can't figure out how to only highlight 1 without some sort of hard-coding

rudouglas:
Ok I think i've got this, we just need to add the `activeSite` parameter to both `MainOverlay` (developer) and `QuickStart` (IO) overlay and this works, but i'm not sure if this was the expected method you were thinking of @zstix 

This PR has the code changes but will fail without the updated theme
https://github.com/newrelic/developer-website/pull/1612
Screenshot 2021-09-13 at 12 33 10
Screenshot 2021-09-13 at 12 33 15

aswanson-nr (approved):

PR #469, opened 2021-09-16
Fix builds by removeing useStaticQuery

zstix:
## Description
We no longer need to query GraphQL for this. This should _hopefully_ fix our issues with this branch not building.

### DIFF
--- a/packages/gatsby-theme-newrelic/src/components/OneTrust.js
+++ b/packages/gatsby-theme-newrelic/src/components/OneTrust.js
@@ -1,30 +1,16 @@
 import React from 'react';
-import { graphql, useStaticQuery } from 'gatsby';
+import PropTypes from 'prop-types'
 
-const OneTrust = () => {
-  const {
-    site: {
-      siteMetadata: { oneTrustID },
-    },
-  } = useStaticQuery(graphql`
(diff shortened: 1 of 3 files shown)


packages/gatsby-theme-newrelic/src/components/OneTrust.js
```diff
@@ -1,40 +1,30 @@
...
-  } = useStaticQuery(graphql`
-    query {
-      site {
-        siteMetadata {
-          oneTrustID
-        }
-      }
-    }
-  `);
+const OneTrust = ({ id }) => {
```
nr-kkenney, line 4: is `id` populated with the same value it was using previously -- `oneTrustID`?

zstix: This is the component prop. I thought it might be a little wonky to say `<OneTrust oneTrustID={oneTrustID} />` but maybe it's better to be consistent.

zstix: Oh, but I forgot to add the prop in. Just updated that.

nr-kkenney (approved):

PR #487, opened 2021-09-29
fix: update swiftype filter for quickstarts

roadlittledawn:
## Descriptions

Update the `type` filter to use what we're setting on the developer site.


jerelmiller (approved):
💯

Issue #489, opened 2021-09-30
[Tessen] Update referrer tracking to capture the full URL  NOT just the path


zoezhou1113:
## Description

This is not a top priority issue. Currently, `context.page.referrer` only captures the path for the internal referrer, and the full URL for the external referrer. The logic should be consistent so we can parse them through the same URL parsing functions.

`note:` this is an issue for all gatsby sites

## Steps to Reproduce
1. Click on other pages on https://developer.newrelic.com/
2. Check the referrer tracking using the Segment Event Tracker extension

## Expected Behaviour
Full URLs are captured for all events

## Additional context
![image](https://user-images.githubusercontent.com/80297589/135461380-73ae1bd4-2505-456d-b3d7-8b07670bcfbf.png)

jpvajda:
@zoezhou1113 is this just an issue on the developer site? Or is this occuring for all our Gatsby sites, I noticed you created this bug in the theme so I was curious

zoezhou1113:
@jpvajda It's on all of them

jpvajda:
👍

jpvajda:
@roadlittledawn the team reviewed this and had some questions: 

- we did work to add referrer url in this [PR ](https://github.com/newrelic/gatsby-theme-newrelic/pull/460) but it's possible we never passed the full path, is that also your thoughts?

roadlittledawn:
@zoezhou1113 i believe the only way we can affect `context.page.referrer` is via setting a directive for the `Referrer-Policy`. we did that previously to set the full URL for referrer when going cross-domain. is there another directive you know of that gets us full URLs when on same domain AND cross domain?

zoezhou1113:
@roadlittledawn I believe for newrelic.com it's just automatically set by Segment: https://segment.com/docs/connections/spec/page/#properties

roadlittledawn:
only other thing we may be able to do is update the `properties.referrer` to the full URL. Kris and I are looking at the analyrics.js code and it seems like we may be overriding the `context.page.referrer` with the value in `properties.referrer`. that didn't seem to be the case with `properties.search` which is a "reserved name" but we set for searches on I/O landing page.

zoezhou1113:
@nr-kkenney This is perfect. Thank you!

PR #492, opened 2021-10-04
chore: update referrer to use full url (#489)

nr-kkenney:
* instead of only grabbing only the pathname, referrer is now set with
the whole url string.

Relates to: #489


roadlittledawn:
this change will only affect data in the `properties.referrer` field and not `context.page.referrer`.

roadlittledawn:
looks like analytics.js was allowing `properties.referrer` to override `context.page.referrer`, so we think this should fix the issue. also worth noting analyrics.js has reserved names for some properties (like `search` which we use in I/O pubilc catalog search) 

https://segment.com/docs/connections/spec/page/#properties

roadlittledawn (approved):

PR #501, opened 2021-10-14
feat: add icon to external links

rudouglas:
Screenshot 2021-10-18 at 12 34 35


jerelmiller:
Should the icon color match the text color?

PR #505, opened 2021-10-18
new 404 page with search functionality

moonlight-komorebi:
## Summary
* when a user lands on a 404 page, the pathname of the url is used to
create a search term which we query swiftype with to display relevant
results to the user. additionally, the results respect the locale of
the user.
* if no results are found, result section of the page is not shown.
* search input is populated with the search term we create mentioned
in the first bullet. when the user clicks the search input, the search
modal opens.
* we track: the url the user tried to navigate when they got a 404,
how many results are shown to the user, clicks on search results
presented to the user.

## Screenshots
Screen Shot 2021-10-18 at 1 57 56 PM
Screen Shot 2021-10-18 at 1 57 28 PM

## Related Links
Resolves: https://github.com/newrelic/docs-website/issues/4183

### DIFF
--- a/packages/gatsby-theme-newrelic/src/pages/404.js
+++ b/packages/gatsby-theme-newrelic/src/pages/404.js
@@ -2,7 +2,6 @@ import React, { useEffect, useState, useCallback } from 'react';
 import PropTypes from 'prop-types';
 import { css } from '@emotion/react';
 import { navigate } from 'gatsby-link';
-import fetch from 'node-fetch';
 
 import GlobalHeader from '../components/GlobalHeader';
 import GlobalFooter from '../components/GlobalFooter';


roadlittledawn:
For reference: I'm using [this set of auto index page URLs](https://gist.github.com/roadlittledawn/5f55d7891d4a82c1053a1bbc6d95cf9b) to test results on the 404 page.

dbarnesbrownNR:
@roadlittledawn and @nr-kkenney, just a couple things:

1. I would use more standard verb and tense than `You may wish` so that it's a little easier to parse for lower proficiency/ESL readers. I'd suggest `You can try again by doing a search:`. 
2. I'd change `Or try one of the following links` to `You can also try one of the following links` to avoid starting the sentence with the capitalized `Or`, which isn't grammatically wrong anymore these days but looks a little weird design-wise due to only being two characters.

LizBaker:
@nr-kkenney when i run it locally and do a search, if i click a result (say for developer) it tries to take me to that path in the url instead of being an external link. does that make sense? i would think i could go to developer/make-a-nerdpack but it goes to localhost/make-a-nerdpack

nr-kkenney:
that was something i asked @roadlittledawn about when we saw it 🙈 

if the url is for developer.newrelic.com, you get routed to localhost because the theme site is running as developer i believe. if you get a result for docs.newrelic, or opensource.newrelic, it should correctly navigate you to those external sites.

PR #535, opened 2021-10-29
Remove external link localization

LizBaker:
This removes the logic which adds a locale to external links to `newrelic.com`. we can't guarantee these paths exist at the moment


roadlittledawn (approved):
:shipit:

PR #539, opened 2021-11-02
Update related resources for quickstarts

LizBaker:
Partially addresses https://github.com/newrelic/developer-website/issues/1692

### DIFF
--- a/packages/gatsby-theme-newrelic/gatsby-node.js
+++ b/packages/gatsby-theme-newrelic/gatsby-node.js
@@ -174,6 +174,22 @@ exports.sourceNodes = (
 
 exports.createResolvers = ({ createResolvers }, themeOptions) => {
   const { layout = {} } = themeOptions;
+  const relatedResources = {
+    args: {
+      limit: {
+        type: 'Int',
+        defaultValue: 5,
+      },
+    },
+    type: ['RelatedResource!'],
+    resolve: (source, args, context) => {
+      const { limit } = args;
+
+      return context.nodeModel
+        .getNodesByIds({ ids: source.children })
+        .slice(0, Math.max(limit, 0));
+    },
+  };
 
   createResolvers({
     Site: {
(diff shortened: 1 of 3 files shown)


packages/gatsby-theme-newrelic/gatsby-node.js
```diff
@@ -216,6 +216,24 @@ exports.createResolvers = ({ createResolvers }, themeOptions) => {
         },
       },
     },
+    Quickstarts: {
+      relatedResources: {
```
zstix, line 220: This block seems pretty similar to lines 202-217...maybe we can move this to a constant? That way we don't need to update things like `defaultValue` in multiple places.

packages/gatsby-theme-newrelic/gatsby-node.js
```diff
@@ -401,8 +419,8 @@ const createRelatedResources = async (
   const { createNode, createParentChildLink } = actions;
 
   if (
-    node.internal.type !== 'Mdx' ||
-    node.fileAbsolutePath.includes(ANNOUNCEMENTS_DIRECTORY)
+    (node.internal.type !== 'Mdx' && node.internal.type !== 'Quickstarts') ||
```
zstix, line 422: Totally a style thing, but a slightly more compact version that might be easier to add onto later could be this:

```suggestion
    !['Mdx', 'Quickstarts'].includes(node.internal.type) ||
```

zstix: If we wanted to be very fancy and abstract it, we could add the option to supply this via theme options. Something like `releatedResourceNodeTypes` (but with a better name). That way we wouldn't have to hard-code a reference to quickstarts here and we could use that for the `createResolvers` bit of this code block too.

```js
const RELATED_RESOURCE_RESOLVER = { /* common settings */ };

createResolvers({
  // ...
  ...themeOptions.relatedResourceNodeTypes.reduce((acc, type) => ({
    ...acc,
    [type]: RELATED_RESOURCE_RESOLVER,
  }), {}),
  // ...
});
```

```js
!['Mdx', ...themeOptions.relatedResourceNodeTypes].includes(node.internal.type) ||
```

To be honest, I'm not sure if that helps, or if it makes it way too complicated. Just trying to explore ways we can keep this abstract since this is pretty specific to only one of our sites.

zstix (commented):
A few abstraction thoughts...please feel free to reject my weird ideas!

packages/gatsby-theme-newrelic/gatsby-node.js
LizBaker: ah yes, that's a good idea!

packages/gatsby-theme-newrelic/gatsby-node.js
LizBaker: i AM a fan of fancy things but i think for this PR i'm going to hold off on that, though i will implement the array check because it's nicer to look at!

zstix (approved):

PR #560, opened 2021-11-12
Clark/documentation update hooks

clarkmcadoo:
I'm probably off on some of these since I don't have great context for their use/formulation, so definitely let me know!

### DIFF
--- a/packages/gatsby-theme-newrelic/README.md
+++ b/packages/gatsby-theme-newrelic/README.md
@@ -89,15 +89,23 @@ websites](https://opensource.newrelic.com).
     - [`useActiveHash`](#useactivehash)
     - [`useClipboard`](#useclipboard)
     - [`useFormattedCode`](#useformattedcode)
+    - [`useHasMounted`](#usehasmounted)
     - [`useInstrumentedData`](#useinstrumenteddata)
     - [`useInstrumentedHandler`](#useinstrumentedhandler)
     - [`useKeyPress`](#usekeypress)
     - [`useLayout`](#uselayout)
+    - [`useLocale`](#uselocale)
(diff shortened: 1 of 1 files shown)


packages/gatsby-theme-newrelic/README.md
```diff
@@ -2653,6 +2661,30 @@ With formatting options:
...
 ```
 
+### `useHasMounted`
+
+A hook that determines if a particular component has been loaded in the DOM.
+
+```js
+import { useHasMounted } from '@newrelic/gatsby-theme-newrelic';
+```
+
```
rudouglas, line 2671: Do we need an `Arguments` header here as well? Even if it's `None`

packages/gatsby-theme-newrelic/README.md
```diff
@@ -2855,6 +2887,68 @@ const MyComponent = () => {
...
+A hook that will get an object of information regarding the local language used within a component.
+
+```js
+import { useLocale } from '@newrelic/gatsby-theme-newrelic';
+```
+
+**Arguments**
+
+n/a
```
rudouglas, line 2900: Wondering if `n/a` means we're not sure or there are none? I see some Arguments also saying `None` or `none` or `There are no arguments for this hook so it would be good to keep this consistent

clarkmcadoo: Yes, i'll shore that up

packages/gatsby-theme-newrelic/README.md
clarkmcadoo: added, but to the line **below** 👇

zstix (approved):

PR #562, opened 2021-11-12
feat: script for publishing each branch as a separate release

rudouglas:
Here is the approach i took. I wrote a Node script we can run locally to publish a new version of the theme to NPM.
The version is specific to the current working branch and also allows for additional versioning e.g

1. I'm working on theme `1.0.0`.
2. I create a new branch called `ru-patch` and make some updates
3. Running the script will create a new version `1.0.0-ru-patch.1` with tag `ru-patch` in NPM
4. I can install this version using the tag/branch name like `yarn add gatsby-theme-newrelic@ru-patch`
4. I make some more updates and run the script again
5. This will create a new version `1.0.0-ru-patch.2` with tag `ru-patch` in NPM
6. Running `yarn add gatsby-theme-newrelic@ru-patch` should pull the latest version `1.0.0-ru-patch.2`

<details>
<summary>Additional Features</summary>

1. Will check that the NPM_TOKEN environment variable has been set
2. Will check you are running Node 16
3. Will check for discrepancies between `yarn.lock` and `package.json`
4. Will check if you are on excluded branches like `main`/`develop`
</details>

You could test this with any repo, but i used this one: https://github.com/rudouglas/gatsby-build-newrelic
All you need is:
 - Create a new publish token in NPM
 - Change the `"name": "ruairi-test"` in `package.json` to something else

### Remaining Questions
I wasn't 100% sure how the current release works, so not sure what happens if I merge a branch with version `1.0.0-ru-patch.1` into main branch with version `1.0.0`, how does semantic versioning deal with that?

### DIFF
--- a/.github/branch-release.js
+++ b/.github/branch-release.js
@@ -1,7 +1,7 @@
 const { execSync } = require("child_process");
 var pjson = require("../packages/gatsby-theme-newrelic/package.json");
 const currentVersion = pjson.version;
-
+const PACKAGE_PATH = './packages/gatsby-theme-newrelic';
 /**
  * Check that the NPM_TOKEN environment variable is set
  * @returns {String} Either the latest version or ''
@@ -84,7 +84,7 @@ const getNextBranchVersion = (branchName, branchTagVersion) => {
  */
 const setYarnVersion = (version) => {
   const yarnVersion = executeCommand(
-    `yarn version --new-version "${version}" --no-git-tag-version`
+    `yarn --cwd ${PACKAGE_PATH} version --new-version "${version}" --no-git-tag-version`
   );
   console.log(`✅ Yarn version set ✅`);
   console.log(yarnVersion);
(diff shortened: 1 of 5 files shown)


.github/branch-release.js
```diff
@@ -0,0 +1,157 @@
...
+  return executeCommand(`git rev-parse --abbrev-ref HEAD`);
+};
+
+/**
+ * Check the latest version on a given branch tag
+ * @param {String} tag The tag to check
+ * @returns {String} Either the latest version or ''
+ */
+const checkBranchTagExists = (tag) => {
+  return executeCommand(`npm view ruairi-test dist-tags.${tag}`);
```
clarkmcadoo, line 35: Might wanna look at replacing `ruairi-test` w/ `@newrelic-...` 👍

jerelmiller:
Since this repo uses [lerna](https://github.com/newrelic/gatsby-theme-newrelic/blob/develop/lerna.json) to create releases, have we thought about just using that tool to handle this? You can use [`lerna publish --canary`](https://github.com/lerna/lerna/tree/main/commands/publish#--canary) to generate `alpha` releases using the git SHA, which I feel like would better match the need to publish releases other than from the release branch.

I know I haven't been part of the discussion, but what is the need to publish new versions from branches instead of the release branch?

rudouglas:
@jerelmiller we did try using lerna to publish feature branches before but ran into issues, I don't think we were using `--canary` though. Looking into how we would configure that now.

The use case here is basically so we can test changes to the theme locally without doing a full release. We took the same approach as Gatsby do, which is to publish a version for every branch we push to make it easier to pull in specific versions to a project

PR #564, opened 2021-11-12
Fix google structured data parsing error

roadlittledawn:
## Description

Updates the google structured data snippet to match the doc on it. The `potentialAction` property is an object and not an array. Google search console sent a notification about the parsing error.

[google sitelinks search box](https://developers.google.com/search/docs/advanced/structured-data/sitelinks-searchbox)


jpvajda (approved):

roadlittledawn:
we think the parsing error came from an older version that was deployed with a trailing comma. just checked google search console and it says it's valid now

PR #566, opened 2021-11-15
feat: replacing pageAction with Tessen

rudouglas:


packages/gatsby-theme-newrelic/package.json
```diff
@@ -7,9 +7,9 @@
     "access": "public"
   },
   "scripts": {
-    "build": "gatsby build",
+    "build": "GATSBY_NEWRELIC_ENV=production gatsby build",
```
nr-kkenney, line 10: why production vs staging?

rudouglas: This is so it reports to the Staging account in NR1 with the rest of the data

nr-kkenney: but to double double check, that is what we want? like if we run this locally, we want that data mixing into the same environment as our prod data. is that 👍 

if so, feel free to resolve this.

packages/gatsby-theme-newrelic/README.md
```diff
@@ -2653,74 +2652,38 @@ With formatting options:
...
-  attached to the `attributes` property of the page action. You can pass a
-  function to instrument dynamic data. If this is a function, the function will
-  be called with the same arguments passed to the handler.
+- `handler` _(function)_: The function hander that should be augmented with Tessen instrumentation. 
+  This can be `null` or `undefined`.
+- `attributes` _(object | function)_: Data passed to the `Tessen.track` API when called. 
+  The attributes **MUST** contain...
+
+  - `tessenEventName` - Needs to be in [Camel Case](https://en.wikipedia.org/wiki/Camel_case)
+  - `tessenCategoryName` - Needs to be in [Title Case](https://en.wikipedia.org/wiki/Title_case)
```
zstix, line 2671: I don't think we need to prepend these variables with `tessen` since the extra context isn't necessary (we're in a tessen track block, we can assume they're for tessen).

Furthermore, according to [the docs](https://pages.datanerd.us/commune/project-tessen/#globally-required-attributes) these variables are `eventName` and `category`. We might want to use those names to stay consistent.

packages/gatsby-theme-newrelic/src/hooks/useInstrumentedHandler.js
```diff
@@ -1,42 +1,49 @@
...
-const useInstrumentedHandler = (handler, attributes, agent = 'newrelic') => {
+const useInstrumentedHandler = (handler, attributes) => {
   const savedHandler = useRef();
   const tessen = useTessen();
 
   useEffect(() => {
     savedHandler.current = handler;
   }, [handler]);
-
+  // console.log({ attributes });
```
nr-kkenney, line 14: we can remove this

packages/gatsby-theme-newrelic/src/pages/404.js
```diff
@@ -15,9 +15,12 @@ import getLocale from '../../gatsby/utils/getLocale';
...
 
-const track = (actionName, attributes = {}) => {
-  if (typeof window !== 'undefined' && window.newrelic && actionName) {
-    window.newrelic.addPageAction(actionName, {
+const track = (trackName, attributes = {}) => {
+  if (typeof window !== 'undefined' && window.Tessen && trackName) {
+    window.Tessen.track({
+      tessenEventName: 'error404',
+      tessenCategoryName: 'ErrorPage',
+      name: trackName,
```
nr-kkenney, line 23: i don't think we need name, trackName. unless '404_redirect' was something we wanted to keep on this new name field.

packages/gatsby-theme-newrelic/src/utils/createTessen.js
```diff
@@ -33,6 +35,24 @@ const tessenAction =
...
+          name
+        )}'.`
+      );
+    }
+
+    if (!TITLE_CASE.test(category)) {
+      return warning(
+        TITLE_CASE.test(category),
+        `tessen.${action}: The 'category' argument needs to be in TitleCase. This has resulted in a noop. Please change '${category}' to something like '${convertToTitleCase(
+          category
```
nr-kkenney, line 51: i like the intent here, but this may be a bit much for just letting us know that our fields are incorrectly formatted. 

a particular edge case we introduce with this is if we accidentally supply null, undefined, or data of an incorrect type, this will throw an error.

we could rip this out, or update the functions to be more robust, but i think as long as the warning tells us the value needs to be TitleCase or CamelCase thats enough.

packages/gatsby-theme-newrelic/src/components/GlobalHeader.js
```diff
@@ -123,11 +123,14 @@ const useSearchQuery = () => {
         setQueryParam('q', searchTerm);
         if (
           typeof window !== 'undefined' &&
-          window.newrelic &&
+          window.Tessen &&
```
LizBaker, line 126: do we need to use tessen off the window here? can we use the tessen hook?

packages/gatsby-theme-newrelic/src/components/NavLink.js
```diff
@@ -101,8 +101,11 @@ const NavLink = ({
             e.preventDefault();
             e.stopPropagation();
             onToggle && onToggle();
-            if (typeof window !== 'undefined' && window.newrelic) {
-              window.newrelic.addPageAction('navInteraction', {
+            if (typeof window !== 'undefined' && window.Tessen) {
```
LizBaker, line 104: same here, hook?

packages/gatsby-theme-newrelic/src/pages/404.js
```diff
@@ -15,9 +15,11 @@ import getLocale from '../../gatsby/utils/getLocale';
 import useThemeTranslation from '../hooks/useThemeTranslation';
 import Trans from '../components/Trans';
 
-const track = (actionName, attributes = {}) => {
-  if (typeof window !== 'undefined' && window.newrelic && actionName) {
-    window.newrelic.addPageAction(actionName, {
+const track = (attributes = {}) => {
+  if (typeof window !== 'undefined' && window.Tessen) {
```
LizBaker, line 19: also here

LizBaker (approved):

Issue #563, opened 2021-11-10
[Theme] Use an anonomous ID to generate the event data we need for conversion funnels.


jpvajda:
### Example query with the anonymous ID

```sql
FROM TessenAction SELECT filter(uniqueCount(anonymousId), WHERE eventName = 'TLMK_SignUp_signup'), filter(uniqueCount(anonymousId), WHERE eventName = 'TLMK_SignUp_signupThankYou') TIMESERIES AUTO SINCE 1 month ago
````

### Acceptance Criteria

- [x] Discuss with PLG how they get the anonymous ID , is it from segment? - `#help-plg` slack channel
- [ ] Determine how we can use this ID to generate the event data we need for conversion funnels.
- [ ] Once confirmed, add the anonymous ID to all events via the theme

rudouglas:
Asking in #help-plg on this, will update with more details

jpvajda:
@rudouglas could you add me to the thread, I'm interested in their reply.

rudouglas:
>The anonymousId comes from Segment—at least on the signup page (which is a repo that PLG built, but is transitioning ownership to #help-signup). On the signup page, [here's the source for how we get anonymousId](https://source.datanerd.us/plg/tillamook/blob/4bd42875fe0c5851f9050ea5b9ef09f2cf82f533/src/utils/tessen.js#L20-L31).
For more info, I'd suggest asking in #project-tessen or #wizards-amplitude to see if someone else knows more.

PR #568, opened 2021-11-15
feat(tessen): add segment anonymous id to tessen calls

roadlittledawn:
## Description

If available reads the `ajs_anonymous_id` cookie set by segment and adds that value as the `anonymousId` property in all tessen calls.

This value is stored with encoded quotes (`%22`), so we use `JSON.parse()` to parse it (similar to the PLG code the ticket linked to. Also did that to `ajs_user_id` just in case (even though that value seems to stored without encoding.

## Related issues / PRs
Closes https://github.com/newrelic/gatsby-theme-newrelic/issues/563


nr-kkenney (approved):

nr-kkenney (changes requested):
Looks good, but didn't see tests were failing 🙈

roadlittledawn:
ah me neither. can fix those up. i think some of the tests have the shape of the tessen properties, so will have to update those

nr-kkenney (approved):

Issue #538, opened 2021-11-02
[Theme] Instrument links in related resources so we can measure clicks


jpvajda:
### Issue

instrument related resources so we know who clicked on what and where and this click can it be tied to a conversion such as install or a sign up of a new user

We currently instrument the clicks with a tessen event called [relatedResourcesLinkClick](https://github.com/newrelic/gatsby-theme-newrelic/blob/bbf50b58854a57af6c2488caf3569901ed1a55fa/packages/gatsby-theme-newrelic/src/components/RelatedResources.js#L91) but do we `DO NOT` know where this linked was clicked and how often?

### Question:

- Do quickstarts surfaced in related resources drive conversions for sign ups or installations?

### Acceptance Criteria 

- [x] The source page URL of where the link was clicked is available in the event
- [x]  The page URL the  user was sent to is available in the event

zstix:
Here's a query to get all the events with the `relatedResourcesLinkClick` interaction type: https://staging.onenr.io/0PLREMKboRa

jpvajda:
I mocked out a dashboard component to visualize this  data when this work is done, `I/O Related Resources Clicks (WIP)` it's here: https://staging.onenr.io/0nVjYbDAqQ0

PR #569, opened 2021-11-16
chore: add referrer url to RelatedResources clicks

tabathadelane:
We were already tracking the URL the user was sent to. 
- `relatedResourcesLinkClick` property: `href`

I added the referrer URL that the user came from. 
- `relatedResourcesLinkClick` property: ~~`referrer`~~ `currentUrl`

closes: https://github.com/newrelic/gatsby-theme-newrelic/issues/538

### DIFF
--- a/demo/src/pages/index.js
+++ b/demo/src/pages/index.js
@@ -11,6 +11,7 @@ import {
   Layout,
   Link,
   PageTools,
+  RelatedResources,
   SearchInput,
   SimpleFeedback,
   Surface,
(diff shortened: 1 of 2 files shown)


roadlittledawn:
[nit] i wonder if we could use a different name than `referrer`? the `PageAction` event uses `currentUrl` if we wanted to stay consistent with the same data that's sent there. to me `referrer` indicates that the user arrived on the current URL from another one, which isn't the case with these `tessen.track()` events. also i think `referrer` is reserved for just that from Segment's point of view.

tabathadelane:
> [nit] i wonder if we could use a different name than `referrer`? the `PageAction` event uses `currentUrl` if we wanted to stay consistent with the same data that's sent there. to me `referrer` indicates that the user arrived on the current URL from another one, which isn't the case with these `tessen.track()` events. also i think `referrer` is reserved for just that from Segment's point of view.

@roadlittledawn That makes sense. That was how it was referenced in the segment extension, so I ran with it. I'll update the code! Thanks for the information. I'm still stumbling through all the jargon.

zstix (approved):

Issue #510, opened 2021-10-20
[Docs] [Docs I/A] Apply design refresh to global header


roadlittledawn:
## Summary

Apply design refresh to global header.

### Design Questions 

- [x] Verify mobile exp.
- [x] confirm dark mode color 
- [x] where do the dark mode / light mode toggle docs 
- [x] What should the language toggle look like?
- [x] What is the final version of the header we need to build?
- [x] Remove adding background color for current site in menu of sites?
- [x] Header mobile nav menu: does the main menu go away or where should it be displayed on mobile? 

## Design

Changes include:
- increase height to 72px (verify with design mock)
- always that darker charcoal color
- search input is lighter so it's more visible
- free account and login button style changes

**Figma**
[Desktop (light mode)](https://www.figma.com/proto/0jsTPLtwz6Bljndkbr1COn/open-source-DS-%5BWIP%5D?page-id=56%3A39509&node-id=150%3A52221&viewport=501%2C48%2C1.05&scaling=min-zoom&starting-point-node-id=150%3A52221&show-proto-sidebar=1)

![2021-11-09_09-16-22](https://user-images.githubusercontent.com/2952843/140973759-3864a432-3c4f-4be8-894f-3be17d397c5a.png)

[Mobile](https://www.figma.com/proto/0jsTPLtwz6Bljndkbr1COn/open-source-DS-%5BWIP%5D?page-id=141%3A43127&node-id=141%3A43617&viewport=501%2C48%2C1.01&scaling=scale-down&starting-point-node-id=141%3A43617)

## Acceptance criteria
- [x] Header height is increased to 72px (verify w/design mock)
- [x] Header color is always dark, charcoal color
- [x] Search input is lighter (per design spec) color, placeholder text updated
- [x] dark/light mode toggle is larger
- [x] Free account and login button styles updated
- [x] Current site background color style changes per spec
- [x] Mobile header updated per design spec

## Helpful links

jpvajda:
@roadlittledawn thought by end of this week, this design may be finalized.

jpvajda:
@roadlittledawn will prepare this ticket for the team

roadlittledawn:
re: main menu in header. should stay the same. scrollable left/right on smaller desktop screens. does not display on mobile. but this is on Àngel's radar for a full fledged redesign of navbar in the future.

legnaleama:
@jpvajda @roadlittledawn A couple of notes on the requirements:

*Buttons*
[...]

PR #570, opened 2021-11-16
style: updated global header to design specs

josephgregoryii:
Styling changes for updated global header designs.

### Desktop
Lightmode

Darkmode

### Mobile

Lightmode

Darkmode

resolves: #510


packages/gatsby-theme-newrelic/src/components/Button.js
```diff
@@ -45,6 +49,14 @@ const styles = {
         color: var(--link-hover-color);
       }
     `,
+    [VARIANTS.LOCALE]: css`
```
jerelmiller, line 52: I'm curious, what is a "locale" button? Is this only for the global header? If so, its worth considering making this a one-off there instead of a global type. Then again I have no context for these changes, so feel free to ignore me 😄

packages/gatsby-theme-newrelic/src/components/GlobalHeader.js
```diff
@@ -304,7 +305,7 @@ const GlobalHeader = ({ className, activeSite }) => {
                 -ms-overflow-style: -ms-autohiding-scrollbar;
 
                 > li {
-                  margin: 0;
+                  margin: 1;
```
jerelmiller, line 308: Since this is now non-zero, this should have a unit associated with it.

packages/gatsby-theme-newrelic/src/components/GlobalHeader.js
```diff
@@ -344,6 +345,7 @@ const GlobalHeader = ({ className, activeSite }) => {
             <li
               css={css`
                 flex: 1;
+                padding: 0px 16px;
```
jerelmiller, line 348:
```suggestion
                padding: 0 1rem;
```

We like to use rems for text. Let's update this to use rems as well (`1rem == 16px`)

packages/gatsby-theme-newrelic/src/components/GlobalHeader.js
```diff
@@ -418,12 +426,20 @@ const GlobalHeader = ({ className, activeSite }) => {
             >
               <Button
                 as={ExternalLink}
-                size={Button.SIZE.EXTRA_SMALL}
+                size={Button.SIZE.SMALL}
                 variant={Button.VARIANT.LINK}
                 href="https://one.newrelic.com"
                 css={css`
+                  margin: 0 10px;
```
jerelmiller, line 433:
```suggestion
                  margin: 0 0.625rem;
```

We like to use `rem` for margins. Let's update to use `rem` here.

packages/gatsby-theme-newrelic/src/components/GlobalNavLink.js
```diff
@@ -35,21 +35,22 @@ const GlobalNavLink = ({ children, href, activeSite }) => {
         display: flex;
         align-items: center;
         height: 100%;
-        padding: 0.25rem 9px;
-        color: var(--secondary-text-color);
-        font-size: 0.6875rem;
+        padding: 4px 9px;
```
jerelmiller, line 38:
```suggestion
        padding: 0.25rem 0.625rem;
```

Huh, not sure why we mixed units here. Could we use rem for padding instead?

packages/gatsby-theme-newrelic/src/components/GlobalNavLink.js
```diff
@@ -35,21 +35,22 @@ const GlobalNavLink = ({ children, href, activeSite }) => {
...
 
-          .dark-mode & {
-            background-color: var(--color-dark-300);
-            color: var(--color-dark-700);
-          }
+        &:active {
+          background-color: var(--color-neutrals-900);
+        }
+
+        :hover {
```
jerelmiller, line 52:
```suggestion
        &:hover {
```

This looks to be missing the `&`

packages/gatsby-theme-newrelic/src/components/NewRelicLogo.js
```diff
@@ -4,7 +4,7 @@ import { css } from '@emotion/react';
 import styled from '@emotion/styled';
 
 const Letter = styled.path`
-  fill: var(--heading-text-color);
+  fill: var(--color-neutrals-050);
```
jerelmiller, line 7: Does this color work for both light and dark mode?

packages/gatsby-theme-newrelic/src/components/SearchInput.js
```diff
@@ -97,18 +98,18 @@ const SearchInput = forwardRef(
           }}
           css={css`
             width: 100%;
-            border: 1px solid var(--border-color);
+            border: 0px solid var(--border-color);
```
jerelmiller, line 101:
```suggestion
            border: 1px solid transparent;
```

I see that a border is added on hover. By reducing this to `0`, I fear this will make the surrounding UI jump since it goes from no border to a 1px border around it. Rather than removing the border in the default state, instead let's change the border color to transparent so that it takes up the same amount of room and avoids jumping around.

packages/gatsby-theme-newrelic/src/components/SearchInput.js
```diff
@@ -97,18 +98,18 @@ const SearchInput = forwardRef(
...
-            line-height: 1;
+            line-height: 1.5rem;
 
             padding-left: calc(
-              var(--horizontal-spacing) + 0.5rem + var(--icon-size)
+              var(--horizontal-spacing) + 0.825rem + var(--icon-size)
             );
             padding-right: calc(
-              var(--horizontal-spacing) + 0.5rem + var(--icon-size)
+              var(--horizontal-spacing) + 0.825rem + var(--icon-size)
```
jerelmiller, line 112: These changes will affect all search inputs. Just to be sure, are you meaning to update this for all search inputs or just the search input in the global header?

packages/gatsby-theme-newrelic/src/components/SearchInput.js
```diff
@@ -97,18 +98,18 @@ const SearchInput = forwardRef(
           }}
           css={css`
             width: 100%;
-            border: 1px solid var(--border-color);
+            border: 0px solid var(--border-color);
             border-radius: 4px;
-            background: var(--primary-background-color);
+            background: var(--color-neutrals-700);
```
jerelmiller, line 103: Does this work for both light and dark modes? This `SearchInput` is used globally, not just the header.

packages/gatsby-theme-newrelic/src/components/SearchInput.js
```diff
@@ -118,6 +119,14 @@ const SearchInput = forwardRef(
               border: 1px solid rgba(0, 126, 138, 0.6);
               box-shadow: 0 0 0 4px rgba(0, 126, 138, 0.1);
             }
+
+            &:hover {
+              border: 1px solid var(--border-color);
```
jerelmiller, line 124:
```suggestion
              border-color: var(--border-color);
```

Assuming the change is made above to use a transparent border color, you can instead just update the border color here.

packages/gatsby-theme-newrelic/src/components/Button.js
josephgregoryii: The "locale" button is intended for the dropdown section in the header for language preference. Assuming that we are going to be adding more languages in the future, there may be multiple options where adding a  "locale" button variant may be beneficial. 

I'm not particularly attached to this idea, so I'm open to suggestions and happy to change this.

packages/gatsby-theme-newrelic/src/components/SearchInput.js
josephgregoryii: @zstix and I were trying to figure out where this was occurring. Thanks for that!

zstix (changes requested):
This looks great! We had a few minor comments.

One other thing, which we spoke out on the call is that the header can get into this weird state at some odd breakpoints:

Screen Shot 2021-11-22 at 11 01 26

packages/gatsby-theme-newrelic/src/components/GlobalHeader.js
```diff
@@ -362,12 +371,49 @@ const GlobalHeader = ({ className, activeSite }) => {
...
+                    height: 40px;
+
+                    .dark-mode & {
+                      --background-color: var(--color-dark-500);
+                    }
+
+                    &:hover {
+                      --hover-border-color: var(--color-neutrals-600);
+                      border-style: solid;
+                      border-color: --hover-border-color;
```
zstix, line 397: I noticed that the placeholder text "wiggles" a little bit when hovering over the input. This is (probably) because the border is pushing text over. If we want to avoid this, we could use an [inset `box-shadow`](https://developer.mozilla.org/en-US/docs/Web/CSS/box-shadow) to achieve the same effect:
```suggestion
                      box-shadow: inset 0 0 0 2px var(--hover-border-color);
}
```

zstix (dismissed):

josephgregoryii:
Sent a link for Àngel's approval. Once reviewed i'll merge once we get the green light.

packages/gatsby-theme-newrelic/src/components/GlobalHeader.js
```diff
@@ -308,7 +312,7 @@ const GlobalHeader = ({ className, activeSite }) => {
                   flex: 0 0 auto;
                 }
 
-                @media screen and (max-width: 545px) {
+                @media screen and (max-width: 700px) {
```
aswanson-nr, line 315:
```suggestion
                @media screen and (max-width: ${MOBILE_BREAKPOINT}) {
```

packages/gatsby-theme-newrelic/src/components/GlobalHeader.js
```diff
@@ -382,11 +422,35 @@ const GlobalHeader = ({ className, activeSite }) => {
               />
             </li>
             {locales.length > 1 && (
-              <li>
+              <li
+                css={css`
+                  @media screen and (max-width: 545px) {
```
aswanson-nr, line 427: I noticed this `545px` value is used in a few places. It would be nice to pop this out to a variable which would make it easier to change later and remove the "magic number"-ness by giving it a name.

packages/gatsby-theme-newrelic/src/components/GlobalHeader.js
```diff
@@ -408,22 +472,40 @@ const GlobalHeader = ({ className, activeSite }) => {
...
                 css={css`
+                  margin: 0 0.625rem;
                   font-weight: 600;
+                  font-size: 0.875rem;
                   white-space: nowrap;
+                  color: var(--color-brand-400);
+                  border: 1px solid var(--color-brand-400);
+                  border-radius: 4px;
+
+                  @media screen and (max-width: 545px) {
```
aswanson-nr, line 506: Here is another place using the `545px` value

packages/gatsby-theme-newrelic/src/components/GlobalHeader.js
```diff
@@ -453,7 +535,16 @@ const GlobalHeader = ({ className, activeSite }) => {
                 display: flex;
               `}
             >
-              <SplitTextButton />
+              <SplitTextButton
+                style={{
```
aswanson-nr, line 539: If possible, we could move this over to the `css` prop provided by emotion. Similar to how we moved the `SearchInput` over.

packages/gatsby-theme-newrelic/src/components/NewRelicLogo.js
```diff
@@ -4,15 +4,21 @@ import { css } from '@emotion/react';
...
-  fill: var(--heading-text-color);
+  fill: var(--text-color);
 `;
 
-const NewRelicLogo = ({ className, size, omitText }) => (
+const NewRelicLogo = ({
+  className,
+  size,
+  omitText,
+  textColor = 'var(--heading-text-color)',
```
aswanson-nr, line 14: I think this is a similar situation where we could make use of the `css` property provided by emotion to do this in a "css"-y way.

aswanson-nr (changes requested):
Looking good! Found a few more spots where the CSS can be pulled into the `css` prop provided by Emotion.

aswanson-nr (approved):
:shipit:

PR #574, opened 2021-11-18
fix!: issues with Tessen

rudouglas:
BREAKING CHANGE: with the function signature changes we made to tessen, code changes will be required in order to use (tessen in) subsequent versions of the theme.


packages/gatsby-theme-newrelic/src/utils/createTessen.js
```diff
@@ -81,8 +145,8 @@ const tessenAction =
   };
 
 const createTessen = (config) => ({
-  page: tessenAction('page', config),
-  track: tessenAction('track', config),
+  page: tessenPage('page', config),
+  track: tessenTrack('track', config),
```
nr-kkenney, line 149: do we need the `page` and `track` arguments now that we have specific functions?

PR #581, opened 2021-11-19
swap eventName & category arguments for tessen tracking in link

nr-kkenney:
* swapped eventName & category arguments in link component since I think
  they are reverse of what we want.

### DIFF
--- a/packages/gatsby-theme-newrelic/src/components/Link.js
+++ b/packages/gatsby-theme-newrelic/src/components/Link.js
@@ -45,14 +45,15 @@ const Link = forwardRef(
     const handleExternalLinkClick = useInstrumentedHandler(onClick, {
       eventName: 'externalLinkClick',
       category: 'LinkClick',
-      origin: 'gatsbyTheme'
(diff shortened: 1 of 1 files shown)


packages/gatsby-theme-newrelic/src/components/Link.js
```diff
@@ -43,15 +43,15 @@ const Link = forwardRef(
     `);
 
     const handleExternalLinkClick = useInstrumentedHandler(onClick, {
-      eventName: 'gatsbyTheme',
-      category: 'ExternalLinkClick',
+      eventName: 'externalLinkClick',
+      category: 'GatsbyTheme',
```
rudouglas, line 47: This is definitely a great change, and the implementation is up for discussion, but I would say the event we are querying in NR1 should not say `gatsbyTheme`, so if we don't include that in the eventName or category it would probably save confusion
```suggestion
      eventName: 'externalLinkClick',
      category: 'LinkClick',
      origin: 'gatsbyTheme'
```

rudouglas (approved):

PR #589, opened 2021-11-23
fix: adjust header buttons in mobile

josephgregoryii:
fixed margin for mobile display (specifically when width is at `375px`)

Co-authored-by: LizBaker <lbaker@newrelic.com>

### DIFF
--- a/demo/src/pages/index.js
+++ b/demo/src/pages/index.js
@@ -243,6 +243,7 @@ nr1 create --type nerdpack --name pageviews-app
           <div
             css={css`
               display: flex;
+              flex-wrap: wrap;
               gap: 1rem;
               margin-bottom: 2rem;
               align-items: start;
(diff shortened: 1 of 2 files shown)


clarkmcadoo:
@josephgregoryii it looks like some of the nav is cut off on the ipad/ipad pro views.  Are we ok with this? I know we're not going to be able to optimize for everything but I was wondering if this had been looked at.  The docs site has a similar thing going on so I'm gonna go ahead and approve this.

clarkmcadoo (dismissed):

josephgregoryii:
@clarkmcadoo brings up a good point. The breakpoint is a little less than the width of the iPad as we can see here:

<details>
<summary>cutoff</summary>

preview

</details>

Do we want to make the breakpoint higher so then the NR logo becomes the dropdown for this width?

cc: @jpvajda @zstix

jpvajda:
> Do we want to make the breakpoint higher so then the NR logo becomes the dropdown for this width?

I'd vote YES to that suggestion... @josephgregoryii

aswanson-nr (approved):

PR #598, opened 2021-12-01
feat: add NoTranslate component and supporting docs

clarkmcadoo:


zstix (commented):
I'm not sure if the ergonomics are better by handling it in this way:

```jsx
<NoTranslate htmlElement="p">This is some text</NoTranslate>

<p className="notranslate">This is some text</p>
```

zstix:
One way we could handle this (in the docs-website repository) is to make it a component that doesn't render anything in the built pages (like [this](https://github.com/newrelic/docs-website/blob/develop/src/components/MDXContainer.js#L22)) and is complete serialized (using a handler [here](https://github.com/newrelic/docs-website/blob/develop/scripts/actions/utils/handlers.js)).

PR #600, opened 2021-12-02
Add option to disable auto localization of internal links

roadlittledawn:
## Description

Adds an option to the `Link` component to disable auto localizing the link's path if on localized version of site. This supports machine translation disclaimer work where we need to add the link to the english version of a page in the disclaimer text at the top of localized pages.

## Related issues / PRs
Unblocks https://github.com/newrelic/docs-website/issues/2537

### DIFF
--- a/packages/gatsby-theme-newrelic/src/components/Link.js
+++ b/packages/gatsby-theme-newrelic/src/components/Link.js
@@ -19,7 +19,14 @@ const isImageLink = (className) => className === 'gatsby-resp-image-link';
 
 const Link = forwardRef(
   (
-    { to, onClick, instrumentation = {}, displayExternalIcon, ...props },
+    {
+      to,
+      onClick,
+      instrumentation = {},
+      displayExternalIcon,
+      shouldAutoLocalize = true,
+      ...props
+    },
     ref
   ) => {
     const locale = useLocale();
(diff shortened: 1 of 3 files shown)


packages/gatsby-theme-newrelic/src/components/Link.js
```diff
@@ -118,10 +125,14 @@ const Link = forwardRef(
...
-          path: forceTrailingSlashes ? addTrailingSlash(to) : to,
-          locale,
-        })}
+        to={
+          shouldAutoLocalize
+            ? localizePath({
+                path: forceTrailingSlashes ? addTrailingSlash(to) : to,
+                locale,
+              })
+            : addTrailingSlash(to)
```
clarkmcadoo, line 134: would we want the same `forcesTrailingSlashes` ternary if `shouldAutoLocalize` is `false`?

roadlittledawn: yeah probably should. just pushed an update. let me know what you think of the name of the const. i'm not in love with it, but it'll work ¯\_(ツ)_/¯

clarkmcadoo (approved):

packages/gatsby-theme-newrelic/src/components/Link.js
clarkmcadoo: looks great!

PR #604, opened 2021-12-08
fix: can't exit search modal on mobile

rudouglas:
I just removed the set heightsbeing applied to the animated Divs when on Mobile, this was preventing the ability to click outside of the SearchModal component to close it because it took up the entire screen


josephgregoryii (approved):
Great fix! Tested on a mobile device using chrome and safari and works as intended.

roadlittledawn:
worked for me too on iOS 15 chrome, safari, firefox. maybe we could merge this and try to release theme again to see if same lerna error persists? @rudouglas

PR #619, opened 2022-01-10
chore: remove swiftype related resources

rudouglas:


zstix:
Relates to https://github.com/newrelic/docs-website/issues/4481

jerelmiller:
By removing this from the theme, you'll also be removing it from the developer site. Is that intentional? I'd prefer removing the config altogether on the docs site instead so that the related resources aren't used. 

Then again, I haven't been involved in conversation, but that linked ticket @zstix linked to is for the docs site.

PR #622, opened 2022-01-18
TSConfig file for theme

josephgregoryii:
## New Dependencies
- Node Types
- React Types
- React-DOM types
- [gatsby-plugin-typegen](https://www.gatsbyjs.com/plugins/gatsby-plugin-typegen/)
- typescript
- ts-loader
- source-map-loader

## Dev Dependenies
- typescript-eslint/parser
- typescript-eslint/eslint-plugin

### Notes:
- When linting via command line, there will be a *ton* of linter warnings. This will slowly decrease as we adapt typescript formatting.
- Linter should work as previously intended. 
- We may also turn `on` some linter options at a later point that were needed to be turned `off` to allow the same rules as previous.

### Suggestions:
- I would suggest sandboxing in the typescript environment and see how the linter feels with your editor.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -17,19 +17,20 @@ module.exports = {
   overrides: [
     {
       files: ['**/*.{ts,tsx}'],
+      parser: '@typescript-eslint/parser',
       parserOptions: {
         project: ['./tsconfig.json'],
       },
     },
   ],
-  parser: '@typescript-eslint/parser',
   parserOptions: {
     ecmaFeatures: {
       jsx: true,
     },
     ecmaVersion: 2018,
     sourceType: 'module',
-    project: ['./tsconfig.json'],
+    tsconfigRootDir: __dirname, // tsconfig.json lies outside of the package
+    project: 'tsconfig.json',
   },
   globals: {
     jest: 'readonly',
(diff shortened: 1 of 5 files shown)


.eslintrc.js
```diff
@@ -3,16 +3,25 @@ module.exports = {
...
     },
   },
+  overrides: [
+    {
+      files: ['**/*.{ts,tsx}'],
+      parserOptions: {
+        project: ['./tsconfig.json'],
+      },
+    },
+  ],
```
josephgregoryii, line 24: Enables typescript linter for javascript (to help with integrating throughout time) and typescript files

.eslintrc.js
```diff
@@ -39,5 +48,7 @@ module.exports = {
         ignoreRestSiblings: false,
       },
     ],
+    '@typescript-eslint/no-var-requires': 'off',
+    '@typescript-eslint/no-empty-function': 'off',
```
josephgregoryii, line 52: disabling these rules allow for the linter not to give any errors, leaving the codebase at the same linting configuration as previous

packages/gatsby-theme-newrelic/gatsby-config.js
```diff
@@ -1,6 +1,7 @@
 module.exports = ({ layout, newrelic, robots = {}, sitemap = true }) => {
   return {
     plugins: [
+      `gatsby-plugin-typegen`,
```
josephgregoryii, line 4: This is to help import gatsby types when working in gatsby components/files

PR #623, opened 2022-01-19
Refactor file system

josephgregoryii:


packages/gatsby-theme-newrelic/gatsby/type-defs.js
```diff
@@ -0,0 +1,12 @@
+const fs = require('fs');
+const path = require('path');
+
+const SCHEMA_CUSTOMIZATION_TYPES = fs
+  .readFileSync(path.resolve(__dirname, './type-defs.graphql'), {
+    encoding: 'utf-8',
+  })
+
+  // converted toString for type-creation in gatsby-node
+  .toString();
```
nr-kkenney, line 10: why is this needed? i can't parse the comment in my head 🙈

josephgregoryii: @nr-kkenney I guess its not, necessarily. I found a ref in the [docs repo](https://github.com/newrelic/docs-website/blob/a2ff68629745ff16ec7a5c050f9f3d6e1ec94ad1/gatsby-node.js#L268-L294) which i thought sort of 'simplified' the approach with removing `fs` from `gatsby-node.js`.

I guess I was thinking that sending over a string, opposed to a buffer

nr-kkenney (approved):

Issue #625, opened 2022-01-19
Fix copyright date


austin-schaefer:
Currently the global footer says `Copyright © 2021 New Relic Inc.`. We should update the date. If we can switch it to just list current year instead of a hardcoded value, that'd be ideal.

Acceptance criteria:

* Update date, ideally not to a hardcoded value
* Bump theme version for docs, dev, and OSS sites to ensure we're on the new date for all sites

PR #626, opened 2022-01-19
Make year dynamic in global footer

LizBaker:
makes the year in the global footer dynamic so we don't have to update it 🎉 

addresses https://github.com/newrelic/gatsby-theme-newrelic/issues/625

Screen Shot 2022-01-19 at 11 42 09 AM


roadlittledawn (approved):
LGTM! :shipit:

PR #643, opened 2022-03-09
fix: Check if an existing page has a locale other than en

LizBaker:
Adding a second locale resulted in an endless loop in the onCreatePage function, this adds a check to stop that loop

### DIFF
--- a/packages/gatsby-theme-newrelic/gatsby-node.js
+++ b/packages/gatsby-theme-newrelic/gatsby-node.js
@@ -12,7 +12,6 @@ const {
 const pageTransforms = require('./gatsby/page-transforms');
 const { getTessenPath } = require('./gatsby/constants');
 const { getFileRelativePath } = require('./gatsby/utils/fs');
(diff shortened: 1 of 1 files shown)


aswanson-nr (approved):

packages/gatsby-theme-newrelic/gatsby-node.js
```diff
@@ -278,22 +277,18 @@ exports.onCreatePage = (helpers, themeOptions) => {
 
   if (
     !transformedPage.path.match(/404/) &&
-    transformedPage.context.fileRelativePath.includes('src/pages/')
+    transformedPage.context.fileRelativePath.includes('src/pages/') &&
+    transformedPage.context.locale === 'en'
```
nr-kkenney, line 281: this is 🧠  right here. much better than includes 🙈

LizBaker: i'm pretty sure you suggested this at some point but i hadn't quite made the connection 😂

PR #649, opened 2022-04-04
fix for tessen not loading

nr-kkenney:
## Summary
Two fixes were made:
* if tessen is unable to be downloaded, we dont want pages to break in the site. Added a check before each instance of usage to ensure tessen exists.
* when using path prefixes, assets are stored underneath the prefix. Currently, the code that adds tessen is looking in a static place. Updated this so that tessen script path incorporates path prefix into src script path.

## Relates to: https://github.com/newrelic/instant-observability-website/issues/99


nr-kkenney:
opening another better PR 🙈

PR #650, opened 2022-04-04
fix for tessen not loading

nr-kkenney:
## Summary
Two fixes were made:
* if tessen is unable to be downloaded, we dont want pages to break in the site. Added a check before each instance of usage to ensure tessen exists.
* when using path prefixes, assets are stored underneath the prefix. Currently, the code that adds tessen is looking in a static place. Updated this so that tessen script path incorporates path prefix into src script path.

## Links
Relates to: https://github.com/newrelic/instant-observability-website/issues/99

### DIFF
--- a/packages/gatsby-theme-newrelic/src/utils/page-tracking/tessen.js
+++ b/packages/gatsby-theme-newrelic/src/utils/page-tracking/tessen.js
@@ -70,26 +70,24 @@ const trackPageView = ({ config, env, location, prevLocation }) => {
 const initializeTessenTracking =
   ({ config, env, location, prevLocation }) =>
   (options = {}) => {
-    if (!initialized) {
-      if (window.Tessen) {
-        initialized = true;
-        const { segmentWriteKey } = config;
-        window.Tessen.load(['Segment', 'NewRelic'], {
-          Segment: {
-            identifiable: true,
-            writeKey:
-              env === 'production' || env === 'prod'
-                ? segmentWriteKey
-                : DEV_SEGMENT_WRITE_KEY,
-            useAmplitudeSessions: true,
-          },
-        });
+    if (window.Tessen && !initialized) {
+      initialized = true;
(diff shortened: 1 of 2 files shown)


packages/gatsby-theme-newrelic/src/utils/createTessen.js
```diff
@@ -58,26 +58,28 @@ const tessenAction =
...
-        location: 'Public',
-        customer_user_id: customerId,
-        anonymousId,
-      },
-      {
-        Segment: {
-          integrations: {
-            All: true,
-          },
+    if (window.Tessen) {
```
jerelmiller, line 61: Looks like this is already covered on line 51-56:

```js

    if (!window.Tessen) {
      return warning(
        false,
        `tessen.${eventName}: You are attempting to use a Tessen action, but Tessen is not available on 'window'. Calls to '${eventName}' will result in a noop.`
      );
    }
```

packages/gatsby-theme-newrelic/src/utils/page-tracking/tessen.js
```diff
@@ -71,23 +71,25 @@ const initializeTessenTracking =
...
-              : DEV_SEGMENT_WRITE_KEY,
-          useAmplitudeSessions: true,
-        },
-      });
-
-      window.Tessen.identify({});
-
-      options.trackPageView &&
-        trackPageView({ config, env, location, prevLocation });
+      if (window.Tessen) {
```
nr-kkenney, line 74: probably just combine the two conditions into one if

packages/gatsby-theme-newrelic/src/utils/createTessen.js
nr-kkenney: good catch 🙈

aswanson-nr (approved):
Tested locally, works with and without a path prefix!

PR #653, opened 2022-04-25
Search term undefined doesn't destroy navigation

caylahamann:
Since `searchTerm` is not a required prop of `Navigation`, we need to anticipate when it might be undefined

### DIFF
--- a/packages/gatsby-theme-newrelic/src/components/Navigation.js
+++ b/packages/gatsby-theme-newrelic/src/components/Navigation.js
@@ -4,7 +4,7 @@ import PropTypes from 'prop-types';
 import NavigationContext from './NavigationContext';
 
 const sanitizeSearchTerm = (searchTerm) =>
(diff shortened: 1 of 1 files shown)


roadlittledawn:
i think one of the issues was when a user entered `%` the page went white, which seems to still happen for the preview build? 

Steps to repro (since direct linking doesn't seem to work)
1. go to https://gatsbythemedemo-caylasearchtermbug.gtsb.io
2. open search and enter `%`
3. observe the blank white screen

LizBaker (approved):

caylahamann:
@roadlittledawn yikes but that confuses me - shouldn't the input still get sanitized? it is just checking if it is undefined

caylahamann:
@roadlittledawn well it is the exact same behavior as before: https://gatsbythemedemo.gtsb.io

roadlittledawn:
gotcha. yeah wasn't 100% sure if this PR was meant to fix that issue or not. carry on 😬

PR #663, opened 2022-05-09
feat(search): conditional placeholder

x8a:
Added a placeholder text for the search in the header for each of the sites: docs, developer, and open source.

Waiting for confirmation on the copy for en, kr, and jp.

### DIFF
--- a/packages/gatsby-theme-newrelic/src/i18n/translations/jp.json
+++ b/packages/gatsby-theme-newrelic/src/i18n/translations/jp.json
@@ -50,7 +50,12 @@
     "ukSlaveryAct": "英国現代奴隷法"
   },
   "searchInput": {
-    "placeholder": "Docs.newrelic.com、Developer.newrelic.com、OpenSource.newrelic.comを検索"
+    "placeholder": {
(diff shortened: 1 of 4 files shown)


packages/gatsby-theme-newrelic/src/i18n/translations/jp.json
```diff
@@ -50,7 +50,7 @@
     "ukSlaveryAct": "英国現代奴隷法"
   },
   "searchInput": {
-    "placeholder": "Docs.newrelic.com、Developer.newrelic.com、OpenSource.newrelic.comを検索"
+    "placeholder": "を検索"
```
caylahamann, line 53: I assume that this is the same as `search docs` in japanese?

caylahamann (approved):

packages/gatsby-theme-newrelic/src/i18n/translations/jp.json
x8a: This is just `Search` for now! Julien will give me the translations once he has them. Both kr and jp keep their former `Search`. 👌🏻

PR #668, opened 2022-05-18
Send filter information to tessen for search events

clarkmcadoo:
### Network Tab Results:

Screen Shot 2022-05-18 at 1 32 21 PM

### NR Results:

Screen Shot 2022-05-18 at 1 44 29 PM


roadlittledawn:
worth noting:
i also added the `GATSBY_NEWRELIC_ENV` to the theme's gatsby site so PR preview builds will send data to NR1 as well

PR #669, opened 2022-05-19
style(GlobalStyles): edited bullet points colors

x8a:


packages/gatsby-theme-newrelic/src/components/GlobalStyles/colors.js
```diff
@@ -7,6 +7,7 @@ export default css`
   --system-text-primary-light: #293338;
   --system-text-secondary-light: #535e65;
   --system-text-secondary-inverted-light: #cdd3d5;
+  --system-text-bullet-points: #767e80;
```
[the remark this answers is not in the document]
clarkmcadoo: yes

clarkmcadoo (dismissed):

packages/gatsby-theme-newrelic/src/components/GlobalStyles/colors.js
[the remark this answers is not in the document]
LizBaker: these names actually come from the one core color palette and I think we should keep with those and either use an existing color here or choose one from there to add

packages/gatsby-theme-newrelic/src/components/GlobalStyles/GlobalStyles.js
```diff
@@ -125,9 +125,16 @@ const GlobalStyles = ({ layout }) => (
       }
 
       ul > li::marker {
-        color: var(--system-text-secondary-inverted-light);
+        color: var(--system-text-bullet-points);
         .dark-mode & {
-          color: var(--system-background-selected-low-contrast-dark);
+          color: var(--system-text-bullet-points-dark);
+        }
+      }
```
LizBaker, line 132: no need to specify dark mode here, we can just use `--primary-bullet-points-color` since it has both a light and dark mode value in themes.js

x8a: It was already like this before, I just changed the colors. Should I change it to `--primary-bullet-points-color` then?

LizBaker (approved):

packages/gatsby-theme-newrelic/src/components/GlobalStyles/GlobalStyles.js
LizBaker: either way! themes.js lets us use a single variable that has both a light mode and dark mode value. we didn't use a theme var for bullet points before so that's why this code had to specify dark mode

PR #670, opened 2022-05-19
Hide logo text at larger screen width

LizBaker:
Screen Shot 2022-05-19 at 11 12 20 AM


packages/gatsby-theme-newrelic/src/components/GlobalHeader.js
```diff
@@ -152,7 +153,7 @@ const GlobalHeader = ({
     }
   `);
 
-  const hideLogoText = useMedia({ maxWidth: '350px' });
+  const hideLogoText = useMedia({ maxWidth: '460px' });
```
clarkmcadoo, line 156: are we trying to keep this in sync with the `LOGO_TEXT_BREAKPOINT` constant? if so, would it make sense to use that here?

LizBaker: hohoho it sure would, lemme fix that

clarkmcadoo (approved):

PR #682, opened 2022-05-24
Sign up modal 

caylahamann:


caylahamann:
Feedback from Libby: 
- Try making the "Free access to all" title a little smaller
- Increase the spacing between the two main elements (sign up and sign up info)
- Increase padding on sign up info card
- Remove the drop shadow from the info card
- Align the check marks with the header
- Change the check mark color to the success green color for all icons
- Change the color of "Forever" to the accent green (we can't use teal anymore :smiling_face_with_tear:)

clarkmcadoo (approved):

PR #687, opened 2022-05-27
Add light box

LizBaker:


packages/gatsby-theme-newrelic/src/components/Lightbox.js
```diff
@@ -0,0 +1,83 @@
...
+  useKeyPress('Escape', () => setLightboxOpen(false), {
+    ignoreTextInput: false,
+  });
+
+  return (
+    <>
+      <button
+        type="button"
+        onClick={() => setLightboxOpen(true)}
+        css={css`
```
roadlittledawn, line 31: do we want to set the background color to transparent instead of that gray color? and also change cursor to something like a magnifying glass on hover?

packages/gatsby-theme-newrelic/src/components/Lightbox.js
```diff
@@ -0,0 +1,83 @@
...
+                      max-height: 90vh;
+                      max-width: 90%;
+                      width: auto;
+                    }
+
+                    .dark-mode & {
+                      background: ${rgba('#3a444b', 0.75)};
+                    }
+                  `}
+                  onClick={() => setLightboxOpen(false)}
```
roadlittledawn, line 68: maybe a question for design, but do we want to add a close button as well?

roadlittledawn (changes requested):
i think on mobile we want to not use lightbox and instead use current behavior which is open image in new tab

packages/gatsby-theme-newrelic/src/components/Lightbox.js
LizBaker: oh my goodness yep, i got rid of the button background in the dev tools but not in the code 🙈

packages/gatsby-theme-newrelic/src/components/Lightbox.js
LizBaker: i'm using the same style as the search modal which doesn't have a close button but we can certainly add it

clarkmcadoo (dismissed):

clarkmcadoo:
Looks great! The only weird thing i noticed was that on desktop it scrolls to the top of the page when hitting `Esc` to get out of the ligthbox, but that can get revised down the line.

clarkmcadoo (approved):

PR #688, opened 2022-06-07
dark mode animation

clarkmcadoo:
https://user-images.githubusercontent.com/26727669/172471685-4b6acf80-569b-4426-aacd-c3d63fd98535.mov


packages/gatsby-theme-newrelic/src/components/DarkModeToggle.js
```diff
@@ -56,11 +56,37 @@ const DarkModeToggle = ({ className, size, onClick }) => {
       `}
     >
       <Icon
-        name={darkMode.value ? 'nr-dark-mode' : 'nr-light-mode'}
+        name="nr-light-mode"
```
caylahamann, line 59: should we change the name of this icon to something like `nr-dark-mode-toggle` or vice versa?

caylahamann (commented):
Looking at the preview, the icon is transposed if you start on dark mode so when you click it doesn't change, and also I'm not seeing the actual animation but I'm really not sure why that is

packages/gatsby-theme-newrelic/src/components/DarkModeToggle.js
clarkmcadoo: we should, its not defined in this repo though i don't think? at least I can't find it

clarkmcadoo: lol it is, i fixed

PR #691, opened 2022-06-13
Add mobile aspect ratio option for videos

LizBaker:
Screen Shot 2022-06-13 at 10 01 38 AM

### DIFF
--- a/demo/src/pages/index.js
+++ b/demo/src/pages/index.js
@@ -421,6 +421,13 @@ nr1 create --type nerdpack --name pageviews-app
           <Video id="inyshp3m7r" type={Video.TYPE.WISTIA} width="500px" />
           <h2>YouTube video</h2>
           <Video id="ZagZfNQYJEU" type={Video.TYPE.YOUTUBE} width="500px" />
+          <h2>Vertical video</h2>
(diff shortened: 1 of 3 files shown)


clarkmcadoo (dismissed):

caylahamann (commented):
i'm not sure if we can do anything about this, but it appears that the control bar is a bit smaller for the vertical view (like in terms of height and font size) 

### Vertical
Screen Shot 2022-06-13 at 2 20 02 PM

### Horizontal

Screen Shot 2022-06-13 at 2 20 09 PM

LizBaker:
> i'm not sure if we can do anything about this, but it appears that the control bar is a bit smaller for the vertical view (like in terms of height and font size)
> 
not seeing anything in my searches so far :[]

roadlittledawn (changes requested):
Can we also update the package [README](https://github.com/newrelic/gatsby-theme-newrelic/blob/liz/aspect-ratio-video/packages/gatsby-theme-newrelic/README.md#video) for the additional prop / usage when it needs to be vertical?

clarkmcadoo (approved):

PR #692, opened 2022-06-14
fix: remove split test text 'Free Account'

clarkmcadoo:
Getting rid of split.io testing of the sign up button text


roadlittledawn (approved):
LGTM

should we remove any of the [split config on docs](https://github.com/newrelic/docs-website/blob/39a33d909e6dba90b42cb2e04e1fe71bfd452104/gatsby-config.js#L112) when we bump theme there? maybe the `features` bit only?

PR #701, opened 2022-06-27
feat(components): skeleton

x8a:
Added a skeleton component so we can use it with Split tests.

### DIFF
--- a/demo/src/pages/index.js
+++ b/demo/src/pages/index.js
@@ -26,6 +26,7 @@ import {
   Lightbox,
 } from '@newrelic/gatsby-theme-newrelic';
 
+import Skeleton from '../../../packages/gatsby-theme-newrelic/src/components/Skeleton';
 import tallImage from '../images/nr-one-ajax-browser.png';
 import regularImage from '../images/apm-intro-overview.png';
 import transparentBackgroundImage from '../images/intro-DT.png';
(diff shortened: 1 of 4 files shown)


packages/gatsby-theme-newrelic/src/components/Skeleton.js
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+import { css } from '@emotion/react';
+
+const Skeleton = () => (
+  <div
+    css={css`
+      width: 100px;
+      height: 100px;
+      animation: skeleton-loading 1s linear infinite alternate;
+      @keyframes skeleton-loading {
```
caylahamann, line 10: this is looking great, one thing that still remains is to account for dark mode. We'll want to add styles for when the dark mode class name is applied to the page. It should look something like this: 
https://github.com/newrelic/gatsby-theme-newrelic/blob/e92c4259148b0349968629029f5a3cbaa41dce7f/packages/gatsby-theme-newrelic/src/components/TextHighlight.js#L25-L27

demo/src/pages/index.js
```diff
@@ -139,6 +140,9 @@ const IndexPage = () => {
             <li>Final list item</li>
           </ul>
           <h2>Search inputs</h2>
```
caylahamann, line 142: small thing, but I notice that in the demo this breaks up the `SearchInput` section. You're going to want to move this `h2` to be right next to the `SearchInput`

caylahamann (changes requested):
Let me know if you have any questions about this feedback!!

caylahamann (approved):
This looks great!!! thanks for making those changes

x8a:
@roadlittledawn, thanks for the notes! I added those 💃🏻.

PR #712, opened 2022-07-06
NR-21573: Add theme color for top notch

mickeyryan42:
https://issues.newrelic.com/browse/NR-21573

Tested on Safari and works.  Need to test on mobile as well.  
Also the ticket leaves open between the greyish background color and brand green.  This implements the background, but looking for opinions between this and the green.

### DIFF
--- a/packages/gatsby-theme-newrelic/src/components/SEO.js
+++ b/packages/gatsby-theme-newrelic/src/components/SEO.js
@@ -94,6 +94,14 @@ const SEO = ({ title, location, type, children }) => {
           />
         );
       })}
+
+      {/* Android */}
+      <meta name="theme-color" content="#1d252c" />
+      {/* Windows */}
+      <meta name="msapplication-navbutton-color" content="#1d252c" />
+      {/* iOS Safari */}
+      <meta name="apple-mobile-web-app-status-bar-style" content="#1d252c" />
+
       {getSwiftypeSiteType() && (
         <meta
           className="swiftype"
--- a/packages/gatsby-theme-newrelic/src/components/SEO.js
+++ b/packages/gatsby-theme-newrelic/src/components/SEO.js
@@ -96,11 +96,11 @@ const SEO = ({ title, location, type, children }) => {
       })}
 
       {/* Android */}
-      <meta name="theme-color" content="#1d252c" />
+      <meta name="theme-color" content="--primary-background-color" />
       {/* Windows */}
-      <meta name="msapplication-navbutton-color" content="#1d252c" />
+      <meta name="msapplication-navbutton-color" content="--primary-background-color" />
       {/* iOS Safari */}
-      <meta name="apple-mobile-web-app-status-bar-style" content="#1d252c" />
+      <meta name="apple-mobile-web-app-status-bar-style" content="--primary-background-color" />
 
       {getSwiftypeSiteType() && (
         <meta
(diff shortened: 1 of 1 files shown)


clarkmcadoo (dismissed):

mickeyryan42:
> Looking on this on IOS Safari, it looks great! The color is a shade or two off from the header background color, so we probably want to adjust that to match. After seeing it live, I don't think the brand green is the way to go, as it would just draw attention away from the header itself.

@aswanson-nr I saw the color difference there, I was a little confused by that.  What I noticed is I/O and the Docs site seem to use this #1d252c as `--brand-grey-black`, but when you load the demo in this repo the header is #2a3338 which is a few shades off.  Do you know which is correct, or where i can find a new relic color scheme to verify which it should be?

clarkmcadoo:
Hey @mickeyryan42!  So here's a good file/directory to ref for some of the styles, esp colors: `packages/gatsby-theme-newrelic/src/components/GlobalStyles/colors.js`

also, in the `theme.js` we're using `--primary-background-color` for all the backgrounds and have light and dark denoted

sorry for the sketchy approve then req for approval, but feel free to ping me for a review as well!

mickeyryan42:
Thanks @clarkmcadoo! Tested using the `--primary-background-color` var and it seems to work, and matches the color.  Let me know what you think of this solution.  As Alec said this will affect more than just the I/O site so we wanted to keep the discussion open.

zstix (dismissed):

clarkmcadoo (dismissed):

packages/gatsby-theme-newrelic/src/components/SEO.js
```diff
@@ -94,6 +94,20 @@ const SEO = ({ title, location, type, children }) => {
           />
         );
       })}
+
+      {/* Android */}
+      <meta name="theme-color" content="--primary-background-color" />
```
LizBaker, line 99: `--primary-background-color` is white in light mode and black in dark mode, so i don't think this color is coming through at all. will this still work if we remove the content property entirely? it feels like its falling back to the header color which is good but if someone looks at this in the future it doesn't make sense

LizBaker: better yet, if the different envs don't fallback the same way we might just want to hardcode the header color. not the best, but safe #293338

packages/gatsby-theme-newrelic/src/components/SEO.js
```diff
@@ -94,6 +94,20 @@ const SEO = ({ title, location, type, children }) => {
...
+      {/* Windows */}
+      <meta
+        name="msapplication-navbutton-color"
+        content="--primary-background-color"
+      />
+      {/* iOS Safari */}
+      <meta
+        name="apple-mobile-web-app-status-bar-style"
+        content="--primary-background-color"
+      />
```
aswanson-nr, line 109: I was digging into these meta tags and it looks like we only need this one if the site can be installed as a progressive web app. Safari can key off of the `theme-color` tag for the notch color. [Docs here](https://developer.apple.com/library/archive/documentation/AppleApplications/Reference/SafariWebContent/ConfiguringWebApplications/ConfiguringWebApplications.html)

packages/gatsby-theme-newrelic/src/components/SEO.js
aswanson-nr: @LizBaker, I got this update and it should be good to go if y'all like it!

clarkmcadoo (approved):

LizBaker (approved):

PR #713, opened 2022-07-07
fix(Dependencies): Security resolutions

KyleGashler:


.gitignore
```diff
@@ -71,4 +71,6 @@ yarn-error.log
 # Yarn Integrity file
 .yarn-integrity
 
-static/
\ No newline at end of file
+static/
+
+.idea
```
clarkmcadoo, line 76: just curious what this is for, local noodling?

clarkmcadoo (approved):

.gitignore
KyleGashler: yes. .idea is generated by any of the jetbrains editors. (I use Webstorm instead of vs code)

roadlittledawn (approved):
LGTM!

PR #719, opened 2022-07-12
Explicitly define the signup schema

aswanson-nr:
# Summary
* Because the signup schema was not explicitly defined, the build was
  throwing errors for configs without the `signup` field because the `SEO` component queries for the field
* This defines that schema and makes it optional, allowing for the query
  to work as normal


aswanson-nr:
To see the error I am referring to, take the `develop` branch, remove the `signup` config from the demo site, then run `yarn start`

LizBaker (approved):

clarkmcadoo (approved):

PR #724, opened 2022-07-13
feat(SideBySide): new component

x8a:
Added a new component for displaying content (text, images, code samples, etc.) in two columns.

### DIFF
--- a/demo/src/pages/index.js
+++ b/demo/src/pages/index.js
@@ -13,6 +13,7 @@ import {
   PageTools,
   RelatedResources,
   SearchInput,
+  SideBySide,
   SimpleFeedback,
   Skeleton,
   Surface,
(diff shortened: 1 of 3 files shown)


demo/src/pages/index.js
```diff
@@ -147,6 +148,21 @@ const IndexPage = () => {
             `}
           />
           <br />
+          <h2>Two columns</h2>
+          <SideBySide>
+            <div>
```
roadlittledawn, line 153: i'm curious why we are using `div` tags as part of this. ideally, we do not want to use `div` tags in MDX files. instead, can we use a child component like `Column`?

x8a: We have to look into this once we do the `MDX` treatment, but also, the theme doesn't have the `Column` component. The goal is that writers don't have to deal with `div` tags at all.

x8a:
This is ready for review. We need to merge this and bump the theme in the docs repo, so we can test it fully.

roadlittledawn (dismissed):
good work! took it for a spin and LGTM! 

adding the `Side` component inside the `SideBySide` parent is going to come in handy i think.

Note: To get this fully supported on docs site PR will need:
* bump theme version (once new version is released)
* update `MDXContainer` to handle the component in MDX files
* add handlers for translating MDX files with this component in it

clarkmcadoo (approved):

PR #725, opened 2022-07-13
Move one trust component to seo

caylahamann:
to make sure that the one trust dialog doesn't pop up on embeds


roadlittledawn (changes requested):
IIRC, the onetrust code needs to execute as soon as possible before other scripts. so as high up in `head` as possible. can we move it up? and ensure it's still the first `script` in `head` on another site like docs?

caylahamann:
@roadlittledawn I'm going to look into if there's a way to still use the `preRenderHtml` hook in the context of the layout that we are in

PR #726, opened 2022-07-15
Custom Input Component - NR-32163

clarkmcadoo:
https://user-images.githubusercontent.com/26727669/179238788-749f62c1-7401-49e5-aecd-530746de0a80.mov

https://issues.newrelic.com/browse/NR-32163


clarkmcadoo:
One question I might have for @caylahamann is, what do we want to do w/ the hover on mobile?  It will naturally just not be available, but is this something we'd like to ensure by hiding?

roadlittledawn (commented):
Looks nifty!

Couple things I noticed:
* may want to ask designer for input re: label background / border on error state
* should we add error message when in error state

LizBaker (approved):
🎉

PR #727, opened 2022-07-19
Add tab component

LizBaker:
This adds a tab component as well as a custom select component from the docs site

Screen Shot 2022-07-19 at 1 28 57 PM
Screen Shot 2022-07-19 at 1 29 12 PM


packages/gatsby-theme-newrelic/src/components/Select.js
```diff
@@ -0,0 +1,92 @@
...
+        width: 100%;
+        font-family: inherit;
+        font-size: inherit;
+        cursor: inherit;
+        line-height: inherit;
+        outline: none;
+        color: var(--primary-text-color);
+
+        option {
+          font: -moz-pull-down-menu;
```
clarkmcadoo, line 63: Do we want this to be our `Söhne` font?

LizBaker: i had to google it but for some reason this allows it to inherit the font 🤷‍♀️ so it should be sohne

clarkmcadoo: yeah, i thought it was a variable we didn't have defined, but def didn't notice the `var( ... )` not bein there

clarkmcadoo (approved):

PR #730, opened 2022-07-20
Add form text editor

LizBaker:
currently using the styling we already have for codeblocks, going to inquire about if we want to change this

Screen Shot 2022-07-20 at 11 16 58 AM
Screen Shot 2022-07-20 at 1 33 47 PM

Screen Shot 2022-07-20 at 1 33 55 PM


roadlittledawn:
Do we want to stack the form on smaller widths to avoid the smush?

LizBaker:
@roadlittledawn whoops, definitely, poor mobile always forgotten

LizBaker:
jammed build

packages/gatsby-theme-newrelic/src/components/InteractiveForm.js
```diff
@@ -0,0 +1,151 @@
...
+import { useDebounce } from 'react-use';
+import { css } from '@emotion/react';
+import { scroller } from 'react-scroll';
+import Link from './Link';
+import Callout from './Callout';
+import CustomTextInput from './CustomTextInput';
+import InteractiveOutput from './InteractiveOutput';
+
+const InteractiveForm = () => {
+  const [appName, setAppName] = useState('My Application');
```
clarkmcadoo, line 11: Do we want to make these placeholder values in the CustomTextInputs vs initial state? I know the designs have it as a fully dark font, but that was before we updated the placeholder styles.

LizBaker: hmm not sure, i kind of like that it matches the config file from the start but we can def change it!

clarkmcadoo (approved):

PR #732, opened 2022-07-21
Add FlipCard component

LizBaker:
https://user-images.githubusercontent.com/39655074/180308997-8b4ff67f-5e31-4d88-bdd7-0ed841e0d6b5.mov

### DIFF
--- a/packages/gatsby-theme-newrelic/src/components/FlipCard.js
+++ b/packages/gatsby-theme-newrelic/src/components/FlipCard.js
@@ -1,0 +1,86 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { css } from '@emotion/react';
+import Surface from './Surface';
+
+const FlipCard = ({ className, flipToBack, children }) => {
+  return (
+    <div
+      css={css`
+        width: 300px;
+        height: 300px;
+        background-color: transparent;
+        perspective: 1000px;
+      `}
+      className={className}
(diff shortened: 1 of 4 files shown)


packages/gatsby-theme-newrelic/src/components/FlipCard.js
```diff
@@ -0,0 +1,86 @@
...
+    {children}
+  </Surface>
+);
+
+FlipCard.Front = CardFront;
+FlipCard.Back = CardBack;
+
+FlipCard.propTypes = {
+  className: PropTypes.string,
+  flipToBack: PropTypes.bool,
```
clarkmcadoo, line 74: Super nit-pick, could we name this variable something like `flipped`?  `flipToBack` feels like a state update vs the state of the component itself.

(i'm still gonna approve)

clarkmcadoo (dismissed):

packages/gatsby-theme-newrelic/src/components/FlipCard.js
LizBaker: ooh yes i was trying to think of the right term

clarkmcadoo:
We might _eventually_ want to have the `Hover` card take a component per side for when/if writers use this, but for now I think it looks awesomeeee

clarkmcadoo (approved):

PR #733, opened 2022-07-21
feat(Select): New Select component added

KyleGashler:
image

image


clarkmcadoo:
This is looking really good!  We'll want to add the carrot at the end to denote to a user that this is a dropdown, I lifted this pseudo element from another component (hehe), let's give it a try!

```js
&::after {
        content: '';
        grid-area: select;
        justify-self: end;
        width: 0.5rem;
        height: 0.25rem;
        clip-path: polygon(100% 0%, 0 0%, 50% 100%);
        background-color: var(--border-color);
      }
`

File: packages/gatsby-theme-newrelic/src/components/SelectInLine.js [hunk 1]
```diff
@@ -0,0 +1,68 @@
...
+
+const Select = ({ disabled, className, label, ...props }) => (
+  <div
+    css={css`
+      margin-top: 16px;
+      position: relative;
+      
+      .dark-mode & {
+        background-color: var(--secondary-background-color);
+      }
```
clarkmcadoo, line 13:
```suggestion
      }
      
      &::after {
        content: '';
        grid-area: select;
        justify-self: end;
        width: 0.5rem;
        height: 0.25rem;
        clip-path: polygon(100% 0%, 0 0%, 50% 100%);
        background-color: var(--border-color);
      }
```

clarkmcadoo (dismissed):

roadlittledawn (approved):

PR #750, opened 2022-08-29
fix: temporarily remove the complex feedback form

clarkmcadoo:
Used git to checkout the SHA before the complex feedback form was created and reset the specific files for the original feedback form to be implemented.


roadlittledawn (approved):
LGTM. thanks for doing that

PR #752, opened 2022-08-30
Add `id` to WalkthroughStep and change h2 from h3

caylahamann:
* `id` passes down to create anchor tags
* use h2 tags to create parity between other pages


packages/gatsby-theme-newrelic/src/components/Walkthrough/WalkthroughStep.js
```diff
@@ -10,6 +10,7 @@ const WalkthroughStep = ({
   number,
   onMouseOver,
   onFocus,
+  id,
```
clarkmcadoo, line 13: is `id` needed for the pageTools to work?

clarkmcadoo (approved):

PR #762, opened 2022-09-08
shouldUpdateScroll Theme Option

clarkmcadoo:
fix: add option to theme to allow listing certain routes that can all)ow scrolling via the gatsby api

### DIFF
--- a/packages/gatsby-theme-newrelic/gatsby/should-update-scroll.js
+++ b/packages/gatsby-theme-newrelic/gatsby/should-update-scroll.js
@@ -11,11 +11,11 @@ const shouldUpdateScroll = (
   // If the 'shouldUpdateScroll' theme option exists and is one of the listed routes
   // OR if a user is changing pages/urls, allow for scrolling to top...
   if (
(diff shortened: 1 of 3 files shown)


LizBaker (dismissed):

packages/gatsby-theme-newrelic/gatsby/should-update-scroll.js
```diff
@@ -0,0 +1,27 @@
...
+  // Ensures no scrolling ever when search modal is open
+  const params = new URLSearchParams(window?.location?.search);
+  if (params.has('q')) {
+    return false;
+  }
+
+  // If the 'shouldUpdateScroll' theme option exists and is one of the listed routes
+  // OR if a user is changing pages/urls, allow for scrolling to top...
+  if (
+    (themeOptions?.shouldUpdateScroll &&
```
caylahamann, line 14: this is such a small point, but could we swap the order of these? It is more likely that the pathname will change than the routes matching, so we would want to check that first since it prevents the routes needing to checked

caylahamann (approved):

PR #765, opened 2022-09-14
Feedback validation

LizBaker:


packages/gatsby-theme-newrelic/src/utils/isValidEmail.js
```diff
@@ -0,0 +1,12 @@
+const isValidInput = (value) => value !== undefined && value.length > 0;
```
caylahamann, line 1: could we rename this to be a bit more specific? like is `isValidString` or `isNotUndefined` just because valid input could be anything in my head

caylahamann (approved):
this looks great!

PR #766, opened 2022-09-14
feat(analytics): add support for site to add optimizely via theme option

roadlittledawn:
## Description

Adds support for a site to add optimizely to it to A/B test, given an optimizely project name slug.

Tested on demo and worked, but removed that so we don't send unnecessary data.

### Site config example

```js
optimizely: {
  projectNameSlug: 'slug_obtained_from_optmizely_ui'
}
```

## Related issues / PRs
Closes https://issues.newrelic.com/browse/NR-44970


roadlittledawn:
putting this on ice cuz not needed currently

PR #772, opened 2022-09-21
Add recaptcha to feedback component

LizBaker:


clarkmcadoo (approved):

roadlittledawn:
are we able to control placement of the recaptcha logo/badge that appears in lower right of page?

i think ideally we'd place it within the form box if we could make it look good

LizBaker:
Screen Shot 2022-09-22 at 1 52 47 PM
recaptcha badge is hidden and the privacy policy (required by google) was added to the form

roadlittledawn:
going to put this here just in case we need to reference google policy on hiding badge / displaying policy. https://developers.google.com/recaptcha/docs/faq#id-like-to-hide-the-recaptcha-badge.-what-is-allowed

PR #789, opened 2022-11-01
Upgrade Lerna to v5

tabathadelane:
Updating Lerna to v5 does not seem to add any breaking changes to our repo. 

I did attempt a dry run by trying some suggestions from this [github issue thread](https://github.com/lerna/lerna/issues/51) by running `npx lerna publish patch --no-git-tag-version --no-push --dist-tag beta` and as far as I can tell it is working as expected. Here is the local output of that command:

Screen Shot 2022-11-01 at 10 27 32 AM
Screen Shot 2022-11-01 at 10 27 13 AM

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
     "jest": "^26.4.0",
     "jest-emotion": "^10.0.32",
     "jest-localstorage-mock": "^2.4.3",
-    "lerna": "^5.0.0",
+    "lerna": "^5.6.2",
     "prettier": "^2.0.5",
     "react-test-renderer": "^16.13.1"
   }
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
     "jest": "^26.4.0",
     "jest-emotion": "^10.0.32",
     "jest-localstorage-mock": "^2.4.3",
-    "lerna": "^3.22.1",
+    "lerna": "^5.0.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -26,7 +26,7 @@
     "jest": "^26.4.0",
     "jest-emotion": "^10.0.32",
     "jest-localstorage-mock": "^2.4.3",
-    "lerna": "^3.22.1",
+    "lerna": "^5.0.0",
```
clarkmcadoo, line 29: Do we need this to be `5.0.0` _exactly_? I can't recall if that was just the minimum version or not.  if not: could we update to the latest v5 version? Latest I see is `5.6.2`, this will have the most up to date support 💯

tabathadelane: Ah, good point. I'll try that really quick and see if that still behaves the same.

clarkmcadoo:
Screen Shot 2022-11-01 at 5 00 24 PM

got this running the command from above, should it look the same as your output?

tabathadelane:
> Screen Shot 2022-11-01 at 5 00 24 PM
> 
> got this running the command from above, should it look the same as your output?

oh yeah, i had to add my branch name temporarily to the lerna.json in the `allowBranch` array on line 9.

clarkmcadoo (approved):

PR #791, opened 2022-11-01
Make documentation JSON file queryable through graphiql

ecchilds:
### Jira Ticket
https://issues.newrelic.com/browse/NR-61546

### Description
Add the documentation Json generated by the `yarn docgen` script to graphiql. This makes use of the `gatsby-transformer-json` plugin to make the Json file queryable.


demo/package.json
```diff
@@ -20,8 +20,9 @@
     "gatsby-remark-images": "^5.7.0",
     "gatsby-source-filesystem": "^3.10.0",
     "gatsby-transformer-sharp": "3.10.0",
+    "gatsby-transformer-json": "",
```
clarkmcadoo, line 23: we might want to put something like "^4.24.0" here instead of leaving it blank.  this will make it better for version control for us, and `4.24.0` is the version that currently gets installed when you run `yarn` so that's the latest version, if there's a major version bump we'll want to manage that

clarkmcadoo: also! once this is updated, you'll want to run `yarn` before pushing as we want to get the updated `yarn.lock` file as well

clarkmcadoo (approved):

PR #797, opened 2022-11-03
feat: Homepage skeleton start

JosieMCantu:
WIP - This PR the skeleton to be filled in
Screen Shot 2022-11-03 at 9 51 54 AM

### DIFF
--- a/demo/src/pages/index.js
+++ b/demo/src/pages/index.js
@@ -1,7 +1,10 @@
 import React from 'react';
 import { css } from '@emotion/react';
+import useDarkMode from 'use-dark-mode';
 
 const Homepage = () => {
+  const darkMode = useDarkMode();
+
   return (
     <div
       css={css`
@@ -68,8 +71,9 @@ const Homepage = () => {
             padding: 3px;
             height: 100%;
             width: 20%;
-            background-color: white;
             border-radius: 5px;
+            background-color: white;
+            box-shadow: ${darkMode.value ? 'none' : 'var( --shadow-1)'};
           `}
         >
           TILE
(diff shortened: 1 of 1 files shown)


demo/src/pages/index.js
```diff
@@ -1,10 +1,98 @@
 import React from 'react';
 
 const Homepage = () => {
+  return (
+    <div style={{ textAlign: 'center', padding: '15px' }}>
```
clarkmcadoo, line 5: Rather than using inline styles this way, could we use the `emotion` css prop?  Its very similar to what you have here, its just the syntax is a little bit different: [example](https://github.com/newrelic/gatsby-theme-newrelic/blob/develop/packages/gatsby-theme-newrelic/src/components/Layout/Layout.js#L14-L30)

that way, we stay consistent with the components' styling and syntax, happy to answer questions if need be!

demo/src/pages/index.js
```diff
@@ -1,10 +1,98 @@
...
+          padding: '0px',
+          textAlign: 'center',
+        }}
+      >
+        <li
+          style={{
+            padding: '3px',
+            height: '100%',
+            width: '20%',
+            backgroundColor: 'whitesmoke',
```
clarkmcadoo, line 53: we might want to look at making these pure white, they get a little lost in the light theme mode

demo/src/pages/index.js
JosieMCantu: Yes, thank you. I've switched over to the emotion css prop.

demo/src/pages/index.js
JosieMCantu: You're right, thank you for the suggestion.

clarkmcadoo (approved):

PR #798, opened 2022-11-03
Properly setting Tessen values from Cookies

zstix:
## Description
Previously, we were using `JSON.parse` after reading the cookie value for `ajs_user_id` and `ajs_anonymous_id`. Since those values are strings already, this was throwing an error:

```
Uncaught SyntaxError: Unexpected non-whitespace character after JSON at position [...]
```

## Related Issue(s)
* NR-54405

### DIFF
--- a/packages/gatsby-theme-newrelic/src/utils/createTessen.js
+++ b/packages/gatsby-theme-newrelic/src/utils/createTessen.js
@@ -55,8 +55,8 @@ const tessenAction =
       );
     }
 
-    const customerId = JSON.parse(Cookies.get('ajs_user_id') || 'null');
-    const anonymousId = JSON.parse(Cookies.get('ajs_anonymous_id') || 'null');
+    const customerId = Cookies.get('ajs_user_id') || null;
+    const anonymousId = Cookies.get('ajs_anonymous_id') || null;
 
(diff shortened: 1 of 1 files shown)


packages/gatsby-theme-newrelic/src/utils/createTessen.js
```diff
@@ -55,8 +55,8 @@ const tessenAction =
       );
     }
 
-    const customerId = JSON.parse(Cookies.get('ajs_user_id') || 'null');
-    const anonymousId = JSON.parse(Cookies.get('ajs_anonymous_id') || 'null');
+    const customerId = Cookies.get('ajs_user_id') || null;
+    const anonymousId = Cookies.get('ajs_anonymous_id') || null;
```
roadlittledawn, line 59: don't think this is a biggie, but this string has encoded quotes in it `%22` which makes querying in NR a little awkward. i can't recall if it always has that escaped quote chars or not.

```
...and anonymousId = '"5188d5e8-e919-4dbe-84d4-17046a7236dc"' since...
```

roadlittledawn (approved):
:shipit:

roadlittledawn (approved):
excellent! thanks!

PR #802, opened 2022-11-04
Kate/layout page

kanderson250:
This PR 

1. Adds a Layout Page demoing the Layout components (Collapser, CollapserGroup, Tabs, Table, Side, SideBySide) and their props. 
2. Removes the root url from the Components Library page in sidenav.json (which previously was redirecting to a nonfunctional page). 
Screen Shot 2022-11-04 at 2 54 36 PM
Screen Shot 2022-11-04 at 2 54 53 PM
Screen Shot 2022-11-04 at 2 55 01 PM


demo/src/pages/layout.js
```diff
@@ -0,0 +1,188 @@
...
+            name
+            required
+            type
+          }
+        }
+      }
+    }
+  }
+`;
+
```
ecchilds, line 173: Sorry to nitpick, but could we add a PropTypes field? My linter yells at me if we don't.
```suggestion

LayoutPage.propTypes = {
  data: PropTypes.object.isRequired,
};

```

kanderson250: This has been updated!

LizBaker (approved):

PR #804, opened 2022-11-07
Add Surfaces page

ecchilds:
This PR adds the surfaces page, and tweaks Surface.js so that it's documentation can be generated.


packages/gatsby-theme-newrelic/src/components/Surface.js
```diff
@@ -30,32 +30,49 @@ const styles = {
   },
 };
 
-const Surface = styled.div`
-  border-radius: 0.25rem;
-  box-shadow: var(--shadow-3);
-  text-decoration: none;
+const Surface = ({ base, interactive, children, className }) => (
```
LizBaker, line 33: we use the `as` prop with surfaces so we'll want to add that in here like we did with the tag component

LizBaker (approved):

PR #806, opened 2022-11-08
Adding utils demo page

mary-martinez:


demo/src/pages/utils-examples.js
```diff
@@ -0,0 +1,117 @@
...
+  const tagData = findComponentData('Tag', componentsData);
+  const tagListData = findComponentData('TagList', componentsData);
+
+  return (
+    <Layout.Main
+      css={css`
+        width: 100%;
+      `}
+    >
+      <Layout.Content>
```
LizBaker, line 32: maybe add a page title?

LizBaker (commented):
this is looking great! we'll want to pull in the latest from feat/update-demo so we can get the nav updates. then you can add this page to the nav (see existing examples)

LizBaker (approved):

PR #813, opened 2022-11-29
Make fileName in CodeBlock dynamic

LizBaker:
This will allow content creators to choose the agent config file name/type that is displayed and downloaded via the modified codeblock


tabathadelane (approved):
love it!

PR #821, opened 2022-12-14
Dedupe gatsby transformer sharp

tabathadelane:
the most recent push has updated the demo site sharp plugins to match the ones in the theme package, and now we have
-  one `gatsby-transformer-sharp` 
- one `gatsby-plugin-sharp`
-  two `sharp` but they are both `0.30.x` instead of also including a `0.29.x` version
     - ~~the best i can guess the reason for this might be is because we have two gatsby versions. the theme package is at `4.8.2` and the demo is at `4.4.0`~~  - i tried bumping the gatsby version and this did not resolve it. 

so hopefully this helps!

Screen Shot 2022-12-15 at 11 27 07 AM
Screen Shot 2022-12-15 at 11 28 54 AM
Screen Shot 2022-12-15 at 11 23 15 AM

### DIFF
--- a/demo/package.json
+++ b/demo/package.json
@@ -15,11 +15,11 @@
     "@newrelic/gatsby-theme-newrelic": "^3.2.0",
     "gatsby": "^4.4.0",
     "gatsby-plugin-mdx": "^2.10.1",
-    "gatsby-plugin-sharp": "4.4.0",
+    "gatsby-plugin-sharp": "4.19.0",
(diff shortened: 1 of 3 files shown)


clarkmcadoo (approved):

LizBaker:
can we bump the dev dependency for gatsby to 4.19.0 while we're in here

LizBaker (approved):

LizBaker (approved):

PR #823, opened 2022-12-19
fix: auto-fill search query from URL

sunnyzanchi:
the filter only ever returned `false` or `undefined`, so all parts of the URL path were getting filtered out and not being used as the query string

when testing this in dev, clicking on the search input caused the browser to navigate to `?q=apache/`, causing the trailing slash to display in the input as well. it looks like this is happening in [gatsby-link](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-link/src/rewrite-link-path.js#L47). in the minified code from the npm package, the check on [this line](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-link/src/rewrite-link-path.js#L11) is being minified to just `true`, so the trailing slash is added. i don't see this behavior in production, so i think this is a separate issue

### DIFF
--- a/packages/gatsby-theme-newrelic/src/pages/404.js
+++ b/packages/gatsby-theme-newrelic/src/pages/404.js
@@ -49,9 +49,7 @@ const NotFoundPage = ({
     (term) => {
       const termsToIgnore = ['docs', pageLocale];
 
-      if (!term || termsToIgnore.includes(term)) {
-        return false;
-      }
+      return term && !termsToIgnore.includes(term);
     },
     [pageLocale]
   );


LizBaker (approved):

roadlittledawn:
when i click in the search input, i can't seemingly search / open the modal at all. i think expected behavior would be clicking in search input would open search model with prepopulated search term so user could perform new search if they want (so navigate to `?q=SEARCH_QUERY`)

current behavior navigates to home page, which is strange and not the desired behavior.

LizBaker:
fwiw this does fix the problem outlined in the ticket (it autofills and swiftype searches with those terms) i feel like this other behavior might be a separate ticket

sunnyzanchi:
> when i click in the search input, i can't seemingly search / open the modal at all. i think expected behavior would be clicking in search input would open search model with prepopulated search term so user could perform new search if they want (so navigate to `?q=SEARCH_QUERY`)

@roadlittledawn so i was originally testing this change in the docs-website repo, which does have the behavior you described. i just tried testing in this repo and i get a runtime error, something to do with an import: `_gatsbyjs_reach_router__WEBPACK_IMPORTED_MODULE_3__.resolve is not a function`. i got the same thing in FF and Chrome so i'm not sure why it was taking you to the homepage 🤷 probably worth making another ticket?

PR #826, opened 2023-01-11
create `useLoggedIn` hook and `LoggedInProvider`

sunnyzanchi:
this PR takes the logic @clarkmcadoo and i wrote for [InstallButton.js](https://github.com/newrelic/instant-observability-website/blob/main/src/components/InstallButton.js) and packages it up in the `useLoggedIn` hook and `LoggedInProvider` for easy reuse

check [instant-observability-website#351](https://github.com/newrelic/instant-observability-website/pull/351) for an example usage

### DIFF
--- a/packages/gatsby-theme-newrelic/src/hooks/useLoggedIn.js
+++ b/packages/gatsby-theme-newrelic/src/hooks/useLoggedIn.js
@@ -58,7 +58,7 @@ export const useLoggedIn = (serviceName) => {
  *
  * @returns {Promise<Boolean>}
  */
-const checkIfUserLoggedIn = (serviceName) =>
+export const checkIfUserLoggedIn = (serviceName) =>
   fetch(NERDGRAPH_URL, {
     method: 'POST',
     credentials: 'include',
@@ -85,7 +85,7 @@ const checkIfUserLoggedIn = (serviceName) =>
  *
  * @returns {Boolean | null} A boolean if the key is set, otherwise returns null.
  */
-const getSavedStatus = () => {
+export const getSavedStatus = () => {
   if (typeof window === 'undefined') return null;
   const status = window.sessionStorage.getItem(SAVED_STATUS_KEY);
   if (status != null) return status === 'true';
--- a/packages/gatsby-theme-newrelic/src/hooks/useLoggedIn.js
+++ b/packages/gatsby-theme-newrelic/src/hooks/useLoggedIn.js
@@ -1,6 +1,7 @@
 import { useContext, useEffect, useState } from 'react';
 import { LoggedInContext } from '../components/LoggedInContext';
 
+const SAVED_STATUS_KEY = 'gatsby-theme-newrelic/logged-in-status';
 const NERDGRAPH_URL = 'https://nerd-graph.service.newrelic.com/graphql';
 
 /**
@@ -24,22 +25,28 @@ const NERDGRAPH_URL = 'https://nerd-graph.service.newrelic.com/graphql';
  * @param {string} serviceName
  */
 export const useLoggedIn = (serviceName) => {
+  const savedStatus = getSavedStatus();
   // if `context` is null, we don't have `LoggedInProvider` as
   // an ancestor, so we'll have to make the call ourself.
   const context = useContext(LoggedInContext);
   const [loading, setLoading] = useState(false);
-  const [loggedIn, setLoggedIn] = useState(false);
+  const [loggedIn, setLoggedIn] = useState(savedStatus);
+  const skipCall = context != null || savedStatus != null;
 
   useEffect(() => {
-    if (context != null) return;
+    if (skipCall) return;
 
     setLoading(true);
     checkIfUserLoggedIn(serviceName)
+      .then((isLoggedIn) => {
+        setSavedStatus(isLoggedIn);
+        return isLoggedIn;
+      })
       .then(setLoggedIn)
       .finally(() => {
         setLoading(false);
       });
-  }, []);
+  }, [serviceName, skipCall]);
 
   if (context != null) return context;
   return { loading, loggedIn };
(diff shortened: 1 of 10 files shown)


packages/gatsby-theme-newrelic/src/hooks/useLoggedIn.js
```diff
@@ -0,0 +1,72 @@
...
+ * @returns {Promise<Boolean>}
+ */
+const checkIfUserLoggedIn = () =>
+  fetch(NERDGRAPH_URL, {
+    method: 'POST',
+    credentials: 'include',
+    redirect: 'error',
+    headers: {
+      'Content-Type': 'application/json',
+      'NewRelic-Requesting-Services': 'io-website',
```
roadlittledawn, line 59: if docs site eventually uses this hook, does this value need to be dynamically set? via a theme option maybe?

sunnyzanchi: ah great catch. yeah, i'll add a way to configure this

packages/gatsby-theme-newrelic/src/utils/createTessen.js
```diff
@@ -68,6 +72,8 @@ const tessenAction =
 
     const customerId = getCookie('ajs_user_id');
     const anonymousId = getCookie('ajs_anonymous_id');
+    const loggedIn =
+      getSavedLoggedInStatus(config.product) ?? (await checkIfUserLoggedIn());
```
sunnyzanchi, line 76: is it fine to use `config.product` here to populate the `NewRelic-Requesting-Services` header?

packages/gatsby-theme-newrelic/src/utils/createTessen.js
```diff
@@ -68,6 +72,8 @@ const tessenAction =
 
     const customerId = getCookie('ajs_user_id');
     const anonymousId = getCookie('ajs_anonymous_id');
+    const loggedIn =
+      getSavedLoggedInStatus() ?? (await checkIfUserLoggedIn(config.product));
```
sunnyzanchi, line 76: is it fine to use config.product here to populate the NewRelic-Requesting-Services header?

sunnyzanchi:
tests are failing with this error

```
ReferenceError: fetch is not defined
    at checkIfUserLoggedIn (/Users/szanchi/code/theme-newrelic/packages/gatsby-theme-newrelic/src/hooks/useLoggedIn.js:90:29)
```

packages/gatsby-theme-newrelic/src/utils/createTessen.js
roadlittledawn: my two cents: i think we should probably add a new theme option explicitly for this. maybe call it something like `nerdgraphRequestingServicesHeader` so it's super clear what it's for and because it may require a specific format (looks like slug, separated by hyphens).

that option would be passed to the theme via a site's `gatsby-config.js` in the `@newrelic/gatsby-theme-newrelic` plugin's `options`. probably a new field there, because `newrelic` is specific to the browser agent configs.

after a bit of digging, i don't _think_ the service key passed along in this header requires any other configuring via service gateway or nerdgraph.

roadlittledawn: if we go that route, we'd also want to add doc on that new option here:

https://github.com/newrelic/gatsby-theme-newrelic/tree/develop/packages/gatsby-theme-newrelic#options

packages/gatsby-theme-newrelic/src/hooks/useLoggedIn.js
```diff
@@ -0,0 +1,108 @@
...
+  const savedStatus = getSavedStatus();
+  // if `context` is null, we don't have `LoggedInProvider` as
+  // an ancestor, so we'll have to make the call ourself.
+  const context = useContext(LoggedInContext);
+  const [loading, setLoading] = useState(false);
+  const [loggedIn, setLoggedIn] = useState(savedStatus);
+  const skipCall = context != null || savedStatus != null;
+
+  useEffect(() => {
+    if (skipCall) return;
```
clarkmcadoo, line 37: I think we will want chat about the `skipCall` some more.  We'll definitely want to put some type of expiration on this cookie, just to be sure this will get removed by some mechanism since logging out of the product won't clear this cookie (unless we collab with them? idk that could be something).

sunnyzanchi: the logged in state is stored in sessionStorage, so it'll get cleared whenever the user closes the current tab. i could add an expiration date to the value stored, but i think it'd be kind of overkill for this. i doubt this data will end up persisting for long enough to cause any issues

clarkmcadoo: ah, no that makes sense then

packages/gatsby-theme-newrelic/src/utils/createTessen.js
sunnyzanchi: done https://github.com/newrelic/gatsby-theme-newrelic/pull/826/files#diff-b0b7cd4f1caa98b028107f2cd31a286d39e7a926edecd9cca4bd3150fa1790e2R597

packages/gatsby-theme-newrelic/src/utils/createTessen.js
```diff
@@ -40,7 +44,7 @@ const canSendAction = ({ config, name, category }) =>
 
 const tessenAction =
   (action, config) =>
-  ({ eventName, category, ...properties }) => {
+  async ({ eventName, category, ...properties }) => {
```
clarkmcadoo, line 47: The test modifications you needed to add, is it due to us making this `async`? reason i ask is i'm wondering if we need to do some promise handling anywhere else in the various repos that would use this 🤷

clarkmcadoo (approved):

roadlittledawn (approved):

PR #832, opened 2023-01-19
Add step components

LizBaker:
you can check them out in an mdx page in the demo at `/steps`

Screen Shot 2023-01-18 at 4 15 05 PM

### DIFF
--- a/packages/gatsby-theme-newrelic/src/components/TutorialSteps/TutorialStep.js
+++ b/packages/gatsby-theme-newrelic/src/components/TutorialSteps/TutorialStep.js
@@ -1,0 +1,101 @@
+import React, { Children } from 'react';
+import PropTypes from 'prop-types';
+import { css } from '@emotion/react';
+import { isMdxType } from '../../utils/mdx';
+
+const ALLOWED_HEADINGS = ['h2', 'h3', 'h4'];
+
+const TutorialStep = ({ children, stepNumber, totalSteps }) => {
+  children = Children.toArray(children);
+
+  const title = ALLOWED_HEADINGS.some((level) => isMdxType(children[0], level))
+    ? children[0]
+    : null;
+
+  const content = title
+    ? children.filter((child) => child !== title)
+    : children;
+
+  return (
+    <div
+      css={css`
+        padding: 1.5rem 0 2rem 0;
+        border-top: 1px solid var(--divider-color);
+
+        &:last-child {
+          border-bottom: 1px solid var(--divider-color);
+        }
+      `}
+    >
+      <StepCounter stepNumber={stepNumber} total={totalSteps} />
+      {title}
+      {content.map((column, idx) => {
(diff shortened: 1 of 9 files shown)


tabathadelane (commented):
:danceydoge::

tabathadelane (dismissed):

roadlittledawn:
should we add docs to [README](https://github.com/newrelic/gatsby-theme-newrelic/blob/liz/add-step-components/packages/gatsby-theme-newrelic/README.md) for `Steps` and `Step`?

roadlittledawn:
when clicking `Copy` for a code block, it seems to temporarily expand the main column? not sure if this is a layout issue with demo or new components ¯\_(ツ)_/¯

https://user-images.githubusercontent.com/2952843/213543690-9a034198-507d-4df4-b292-4d8c84f11c61.mp4

packages/gatsby-theme-newrelic/src/components/TutorialSteps/TutorialStep.js
```diff
@@ -0,0 +1,101 @@
...
+import { isMdxType } from '../../utils/mdx';
+
+const ALLOWED_HEADINGS = ['h2', 'h3', 'h4'];
+
+const TutorialStep = ({ children, stepNumber, totalSteps }) => {
+  children = Children.toArray(children);
+
+  const title = ALLOWED_HEADINGS.some((level) => isMdxType(children[0], level))
+    ? children[0]
+    : null;
```
roadlittledawn, line 13: for some reason i'm not super clear on what we're doing here or if it's necessary? if anything, might be worth documenting for writers if they want to set a section title.

if the immediate child of the `Step` component is a `<h2>`, `<h3>`, `<h4>` set it as the title of the section. if not, do nothing and move on.

weirdly, if you use markdown syntax for immediate child like so:

```
<Step>
## Heading 2

More content
```

results in this:



but doing this:

```
<Step>
<h2>Heading 2</h2>

More content
```

results in this, which is more like what's expected?

clarkmcadoo:
> when clicking `Copy` for a code block, it seems to temporarily expand the main column? not sure if this is a layout issue with demo or new components ¯_(ツ)_/¯
> 
>  2023-01-19_11-40-48.mp4

I _think_ this is just how that button and spacing is being handled, it doesn't occur on the lower code block

clarkmcadoo (dismissed):

roadlittledawn (dismissed):
Minor comments. We probably want to at least add doc to README

LizBaker:
for the copy button situation i did find a little bug in SideBySide which i can fix but otherwise unsure if the steps have anything to do with that

as far as refactoring the existing code goes we can do that now but it will slow us down and it means transferring the docs might not be 1:1 copy/paste for the writers 🤷‍♀️ not sure what the best path is

clarkmcadoo (approved):

PR #835, opened 2023-01-23
fix: get deepest descendent for mobile tabs

sunnyzanchi:
trying to render `children` didn't work for mobile tabs from MDX pages because the text we need to render is nested two components deep. this change traverses the children of the component passed in until we get to an element that has no children and renders that, which for tabs is just a string


LizBaker (approved):
we only want the deepest of children here

sunnyzanchi:
@LizBaker i removed a bit of debugging code i accidentally left in the commit, it stale-ized your review but nothing of consequence actually changed

tabathadelane (approved):

PR #842, opened 2023-01-26
Update tab styling options

LizBaker:
This PR is primarily to add a class to the currently selected tab so we can access styling on selected tabs from outside the theme. It also passes css through the page component just in case and updates eslint so it stops yelling at me about emotion's css prop

styling the selected tab from within the demo page
Screen Shot 2023-01-26 at 10 20 39 AM


packages/gatsby-theme-newrelic/src/components/Tabs/BarItem.js
```diff
@@ -39,17 +39,16 @@ const BarItem = ({
...
           border-bottom: var(--brand-button-primary-accent) solid 3px;
 
           .dark-mode & {
             border-bottom: var(--brand-button-primary-accent-hover) solid 3px;
           }
-        `}
+        }
       `}
-      className={className}
+      className={`${className || ''}${isSelected ? ' isSelected' : ''}`}
```
sunnyzanchi, line 51: what do you think about using [`classnames`](https://www.npmjs.com/package/classnames) for this? it's [v small](https://bundlephobia.com/package/classnames@2.3.2)

sunnyzanchi (approved):
left a smol suggestion but by no means a blocker

packages/gatsby-theme-newrelic/src/components/Tabs/BarItem.js
LizBaker: I was thinking about classnames while i did this! 🙈 v small is true, let's bring it in

clarkmcadoo (approved):

sunnyzanchi (approved):

PR #843, opened 2023-01-26
fix: add useEffect to Tabs component

tabathadelane:
tabState was not being updated when initialTab prop was being updated


sunnyzanchi (commented):
just to double check if i understand this, is this so if you're using `<Tabs />`, you can change the current tab from the parent by changing what you pass in for `initialTab`?

like this?
```jsx
const Component = () => {
  const [someId, setSomeId] = useState('some-id')
  useEffect(() => {
    // do something here that would change `someId`
  }, [])

  return <Tabs initialTab={someId} />
}
```

PR #850, opened 2023-02-01
port over useDarkMode hook

LizBaker:
When we upgraded to gatsby v4.25.2 something about the [useDarkMode](https://github.com/donavon/use-dark-mode) hook started causing the darkMode value in localStorage to be set to undefined. It's still not completely clear to me why this happened but in copying the code to our repo to dig in further I found it working as-is. Maybe a conflicting package version? Regardless, the package appears to no longer be maintained so it may be best for us to copy the code to a place we can edit it if needed. I included the original license in the folder with the copied content

### DIFF
--- a/demo/package.json
+++ b/demo/package.json
@@ -13,7 +13,7 @@
     "@mdx-js/mdx": "^1.6.16",
     "@mdx-js/react": "^1.6.16",
     "@newrelic/gatsby-theme-newrelic": "^3.2.0",
-    "gatsby": "^4.4.0",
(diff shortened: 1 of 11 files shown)


clarkmcadoo (approved):

clarkmcadoo:
lgtm - i just clicked the toggle and did a search and the value in localStorage looked 👍 , hopefully that's sufficient QA 🤷

tabathadelane (approved):
^Same. I tried in all three major browsers.

PR #857, opened 2023-02-06
fix: use `visibility` instead of `overflow`

sunnyzanchi:
this allows the `LicenseKey` component to be used within `Collapser`s. previously, the `LicenseKey` popover would be cut off, because it couldn't overflow the container

it doesn't look quite as good, but it does allow the license key popover to display normally

### before

before

### after
after


tabathadelane (approved):

tabathadelane:
I have no strong feelings about the visuals. It's not as great, but if there isn't a simple, fast alternative, I think we want to prioritize getting this fixed in production.

clarkmcadoo (approved):

PR #869, opened 2023-02-13
create InlineSignup component

sunnyzanchi:
reCaptcha doesn't work locally, so after this PR is merged i'll do a theme bump and test this on staging for the docs repo

## screenshots

Screen Shot 2023-02-13 at 3 26 02 PM

Screen Shot 2023-02-13 at 3 26 42 PM

### DIFF
--- a/packages/gatsby-theme-newrelic/src/components/InlineSignup.js
+++ b/packages/gatsby-theme-newrelic/src/components/InlineSignup.js
@@ -1,4 +1,5 @@
 import React, { useState } from 'react';
+import { css } from '@emotion/react';
 import styled from '@emotion/styled';
 import cx from 'classnames';
 
@@ -8,6 +9,7 @@ import useThemeTranslation from '../hooks/useThemeTranslation';
 import Button from './Button';
 import Trans from './Trans';
 import { createAccountRequest } from './SignupModal/signup';
+import RecaptchaFooter from './SignupModal/RecaptchaFooter';
 
 const CTAButton = styled(Button)`
   align-self: start;
@@ -91,7 +93,7 @@ const Terms = styled.p`
   grid-column: 1 / 4;
   font-size: 0.875rem;
   line-height: 1.25;
-  margin-top: 0.25rem;
+  margin: 0.25rem 0 0.5rem;
 
   a {
     color: currentColor;
@@ -196,6 +198,11 @@ const InlineSignup = () => {
           </a>
         </Trans>
       </Terms>
+      <RecaptchaFooter
+        css={css`
+          grid-column: 1 / 4;
+        `}
+      />
     </Form>
   );
 };
--- a/packages/gatsby-theme-newrelic/src/components/InlineSignup.js
+++ b/packages/gatsby-theme-newrelic/src/components/InlineSignup.js
@@ -1,0 +1,203 @@
+import React, { useState } from 'react';
+import styled from '@emotion/styled';
+import cx from 'classnames';
+
+import { useLoggedIn } from '../hooks/useLoggedIn';
+import useTessen from '../hooks/useTessen';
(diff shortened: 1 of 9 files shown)


packages/gatsby-theme-newrelic/src/components/InlineSignup.js
```diff
@@ -0,0 +1,203 @@
...
+
+    if (organizationId) {
+      const redirectUrl = `https://newrelic.com/thank-you?org=${organizationId}&standalone=true`;
+      window.location.assign(redirectUrl);
+    } else {
+      setError(true);
+    }
+  };
+
+  if (loggedIn == null || loggedIn) return null;
```
clarkmcadoo, line 132: is this meant to be _not_ logic?
```suggestion
  if (loggedIn == null || !loggedIn) return null;
```

packages/gatsby-theme-newrelic/src/components/InlineSignup.js
```diff
@@ -0,0 +1,203 @@
...
```
clarkmcadoo, line 132: is there a chance `loggedIn` could ever be 'false'?

packages/gatsby-theme-newrelic/src/components/InlineSignup.js
```diff
@@ -0,0 +1,203 @@
...
+      </InputContainer>
+      <InputContainer>
+        <label className="screenreader-only" htmlFor="inline-signup-email">
+          {t('inlineSignup.emailLabel')}
+        </label>
+        <Input
+          className="last"
+          id="inline-signup-email"
+          name="email"
+          pattern=".+@.+\..+"
```
clarkmcadoo, line 173: the validation logic for the Feedback form doesn't require users to list a top level domain for the email.  We went with this to emulate the signup form on newrelic.com/signup

clarkmcadoo: well nvm, it does required a top level domain now, so we should update the feedback form. diff PR

sunnyzanchi: oh interesting. i'm kind of surprised it ever allowed an email without a TLD 😮. i can update the validation in Feedback in a different change

packages/gatsby-theme-newrelic/src/components/InlineSignup.js
sunnyzanchi: yep, that's specifically the only case we want to show the form. i checked against null because i don't think we want to show the form until we know for sure the user isn't logged in

packages/gatsby-theme-newrelic/src/components/InlineSignup.js
sunnyzanchi: opened #870 for this

packages/gatsby-theme-newrelic/src/components/InlineSignup.js
clarkmcadoo: won't it come in as a string and be truthy?

sunnyzanchi: nope, the hook takes care of any casting if it has to read from sessionStorage, it only ever returns a boolean or null

clarkmcadoo (dismissed):

tabathadelane:
I think we need to add the reCapcha fine print under this form. I am confirming this.

LizBaker:
@sunnyzanchi there's a lil component called RecaptchaFooter that you can pop in there

sunnyzanchi:
@tabathadelane updated with the RecaptchaFooter component

Screen Shot 2023-02-14 at 12 37 45 PM

LizBaker (approved):

clarkmcadoo (approved):

PR #887, opened 2023-03-01
feat: generalize the LicenseKey Popover

sunnyzanchi:
creates two new components, Popover and PopoverButton, that work like the LicenseKey component except they can accept arbitrary children

## testing

- check custom popover at the bottom of the demo page


packages/gatsby-theme-newrelic/src/components/Popover/PopoverButton.js
```diff
@@ -0,0 +1,142 @@
+import React, { useCallback, useEffect, useRef, useState } from 'react';
+import PropTypes from 'prop-types';
+import { css } from '@emotion/react';
+import { Icon, useTessen } from '@newrelic/gatsby-theme-newrelic';
```
LizBaker, line 4: i think we can just directly import these, since we're in the package, right?

sunnyzanchi: yep, updated!

LizBaker (approved):
💖

clarkmcadoo:
see a couple issues here :( 

https://user-images.githubusercontent.com/26727669/223167598-4021ce62-6b2e-4048-9079-606213d0ac5b.mov

clarkmcadoo (approved):

sunnyzanchi:
> see a couple issues here :(

for posterity, this was on the old `sunny/inline-testing` branch, no longer an issue

clarkmcadoo (approved):

PR #888, opened 2023-03-02
Make Link component dynamic based on loggedIn

clarkmcadoo:
Added the LoggedInProvider to list of items being exported from the theme and added that provider to the demo page.

Updated Link component to see if urls are product URLs, if they are, and the user is not logged it, it will send them to the signup page.  If they are logged in, it will send them to the product.

### DIFF
--- a/packages/gatsby-theme-newrelic/src/components/Link.js
+++ b/packages/gatsby-theme-newrelic/src/components/Link.js
@@ -8,7 +8,7 @@ import SignUpLink from './SignUpLink';
 import Icon from './Icon';
 import useInstrumentedHandler from '../hooks/useInstrumentedHandler';
 import { css } from '@emotion/react';
-import useLoggedIn from '../hooks/useLoggedIn';
+import { useLoggedIn } from '../hooks/useLoggedIn';
 
 const isHash = (to) => to.startsWith('#');
 const isExternal = (to) => to.startsWith('http');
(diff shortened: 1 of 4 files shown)


packages/gatsby-theme-newrelic/src/components/Link.js
```diff
@@ -46,6 +47,16 @@ const Link = forwardRef(
       }
     `);
 
+    // Supplied from LoggedInProvider
+    const { loggedIn } = useLoggedIn();
+
+    if (to.includes('one.newrelic.com')) {
+      // Using session storage, so the value will come as a string.
+      if (loggedIn === 'false') {
```
LizBaker, line 55: i think the useLoggedIn hook always returns a boolean and not a string

clarkmcadoo: i was having a lot of trouble getting it to work just via a Boolean and idk why. i'll try again

LizBaker (approved):

packages/gatsby-theme-newrelic/src/hooks/useLoggedIn.js
```diff
@@ -55,8 +55,6 @@ export const useLoggedIn = () => {
   return { loading, loggedIn };
 };
 
-export default useLoggedIn;
```
sunnyzanchi, line 58: i missed this the first time but `gatsby-theme-newrelic/index.js` reexports the default export from this file as `useLoggedIn`. i pushed `71e5a45` to fix this

sunnyzanchi (approved):
✨

PR #906, opened 2023-03-20
Add Osano / remove OneTrust

LizBaker:
this removes onetrust and adds osano. It also adds some logic to add a language so the consent manager language changes with the language picker. since the language picker doesn't fully reload the page it wasn't picking up on the change, i swapped out the gatsby Link for an a tag and it now loads in fresh and picks up the language. 

to test locally you can add ENVIRONMENT=production to the demo develop command. the text that triggers the consent manager is in the footer (see video)
this should not be in the header tag/footer component for the preview as it doesn't have a production env var

https://user-images.githubusercontent.com/39655074/226483648-82696178-7167-43fd-8013-f51bd64099b6.mov


tabathadelane (approved):
I can access the settings and it updates with the language picker.  I dont have that cute cookie though. 🍪 
Screen Shot 2023-03-22 at 4 05 48 PM

packages/gatsby-theme-newrelic/gatsby/on-pre-render-html.js
```diff
@@ -3,22 +3,37 @@ import path from 'path';
...
-import OneTrust from '../src/components/OneTrust';
 
 const onPreRenderHTML = (
-  { getHeadComponents, replaceHeadComponents },
+  { getHeadComponents, replaceHeadComponents, pathname },
   themeOptions
 ) => {
   const tessen = getTessenConfig(themeOptions);
-  const { oneTrustID } = themeOptions;
+  const languages = { kr: 'ko', jp: 'ja', en: 'en' };
```
clarkmcadoo, line 12: `nit` we can probably get rid of 'en' since it'll never get used, right?

clarkmcadoo (dismissed):

packages/gatsby-theme-newrelic/gatsby/on-pre-render-html.js
LizBaker: ahh yes good point

PR #920, opened 2023-04-14
fix: add overflow: scroll 

clarkmcadoo:
Tabs content was bleeding into other page content, this will prevent that from occurring.


tabathadelane:
Oh boy, this brings back the obnoxious scroll bars. They currently aren't there. https://gatsbythemedemo.gatsbyjs.io/ They are so obvious in dark mode. 

Screen Shot 2023-04-14 at 2 07 18 PM

PR #925, opened 2023-04-19
fix Popover offset not working

sunnyzanchi:
the `useCallback` didn't have `show` as a dependency, so it wasn't recalculating the offset correctly. this fixes that and adds `overflow-x: hidden` to the body so that overflowing Popovers don't increase the document size


tabathadelane (approved):
appears to be working as expected
Screen Shot 2023-04-19 at 9 44 25 AM

PR #927, opened 2023-04-19
fix: add `height: 100%` to fix overflow

sunnyzanchi:
## testing

- check out this branch in the theme repo
- in the docs repo:
  - run `cp -r <theme-repo-dir>/packages/gatsby-theme-newrelic/src/components/Tabs/ node_modules/@newrelic/gatsby-theme-newrelic/src/components/Tabs/`
  - `yarn clean`
  - `BUILD_FOLDERS=synthetics yarn start`
  - go to http://localhost:8000/docs/synthetics/synthetic-monitoring/getting-started/get-started-synthetic-monitoring/

### DIFF
--- a/packages/gatsby-theme-newrelic/src/components/Tabs/Bar.js
+++ b/packages/gatsby-theme-newrelic/src/components/Tabs/Bar.js
@@ -35,7 +35,7 @@ const getDeepestChild = (child) => {
 };
 
 const MobileTabControl = ({ children, className }) => {
-  const [[currentTab, setCurrentTab]] = useTabs();
+  const { currentTab, setCurrentTab } = useTabs();
 
   // eslint gets angry about using props from React.Children.map
   /* eslint-disable react/prop-types */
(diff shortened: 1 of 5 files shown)


tabathadelane:
i cannot reproduce this on the site.
Screen Shot 2023-04-19 at 1 34 36 PM

sunnyzanchi:
> i cannot reproduce this on the site.

whoops, sorry, that's my bad. i didn't realize there was a different class on tabpanels that were opened than ones that are closed. the class is actually `css-id30ng`

clarkmcadoo (approved):

clarkmcadoo:
I think the problem here is, we're expecting the overall tabs component to be the height of the largest page up to 500px.  The content causing the issue here is going to 500px and can still scroll, but the Tabs aren't compensating for it.  Maybe we just pass the height all the way back to the Tabs component and set the overall height there?

tabathadelane:
i've followed the steps twice and im still getting this 
Screen Shot 2023-04-20 at 12 00 54 PM

tabathadelane (approved):
it seems to be working as expected when i copy the files over into the node modules
Screen Shot 2023-04-20 at 12 15 45 PM

PR #929, opened 2023-04-25
Use menuItem as a tag

LizBaker:
This makes the entire menu item area clickable instead of just the text


tabathadelane (dismissed):
Seems to work as expected

tabathadelane (approved):

clarkmcadoo (approved):

PR #931, opened 2023-05-04
Add apache to languages

brnhensley:
We have a couple PHP docs using apache syntax, which works on github, so I know it's a valid language identifier

Here's an example:
- https://github.com/newrelic/docs-website/blob/develop/src/content/docs/apm/agents/php-agent/configuration/php-directory-ini-settings.mdx

related PR adding apache code blocks:
- https://github.com/newrelic/docs-website/pull/12932


brnhensley:
I'll update the other docs to use `apacheconf` instead of `apache`

PR #936, opened 2023-05-26
add `sideEffects: false` for tree-shaking

sunnyzanchi:
on the opensource site, this reduces the build output by about a third (3.27MB before vs 1.91MB after)


tabathadelane (approved):
neat!

PR #944, opened 2023-08-31
fix: images overflowing main body container

sunnyzanchi:


tabathadelane (approved):
Copied the changes to the docs node modules and it seems to be working as expected

with changes:
Screenshot 2023-08-31 at 1 26 45 PM

current docs site:
these images are also being stretched off the page in mobile
Screenshot 2023-08-31 at 1 26 40 PM

LizBaker (approved):

PR #957, opened 2023-11-07
feat: update gatsby plugin newrelic

clarkmcadoo:
Screenshot 2023-11-07 at 10 19 15 AM

Updates `gatsby-plugin-newrelic` to use a version that will install the latest NR browser agent


tabathadelane (approved):
🥳

PR #959, opened 2023-12-04
revert Gatsby 5 changes

sunnyzanchi:
this is a lot of changed files, but primarily it just reverts a bunch of commits so that the codebase is back to its state @ 7.1.3, before we made any changes for Gatsby 5. there are some non-Gatsby 5 changes — like removing the AnimatedCard component — that i left in

## testing

i'd recommend clearing `node_modules` first, then `yarn`ing and `yarn start`ing to test the site. the best way to review this PR is probably not to look at the commits or files changed, but to look at the diff between this branch and the `v7.1.3` tag. you can do this after checking out this branch with `yarn diff v7.1.3`. the changelog is huge, you can exclude it with `git diff v7.1.3 ':!*.md'`

things to look out for in the diff:

- any import/require changes, especially anything using `.mjs`
- any dependency changes in `package.json`, `demo/package.json`, or `packages/gatsby-theme-newrelic/package.json`
- any GraphQL changes


.github/workflows/pr_merged.yml
```diff
@@ -23,7 +23,7 @@ jobs:
       - name: Setup Node.js
         uses: actions/setup-node@v3
         with:
-          node-version: 18
+          node-version: 17
```
clarkmcadoo, line 26: thinking we should make this 16

tabathadelane (commented):
No issues stand out. The demo js and mdx pages are working

tabathadelane (approved):

clarkmcadoo (approved):

PR #961, opened 2023-12-07
feat: add privacy icon

sunnyzanchi:
i added a conditional max-width, because otherwise the last item in the list of links is on a line by itself and it looks weird at larger font sizes/smaller screen sizes

### with Your Privacy Choices
Screenshot 2023-12-07 at 2 44 00 PM

### without
Screenshot 2023-12-07 at 2 45 24 PM

## testing
the theme demo doesn't have Osano, so the link we're adding the icon to doesn't show up normally. in GlobalFooter.js, change the `hasOsano` function to always return `true`. then the "Your Privacy Choices" link and the new icon will show up

the behavior of the icon (invert on hover) should match the behavior on newrelic.com. clicking the link will throw an error since Osano isn't defined

### DIFF
--- a/packages/gatsby-theme-newrelic/src/components/GlobalFooter.js
+++ b/packages/gatsby-theme-newrelic/src/components/GlobalFooter.js
@@ -4,6 +4,7 @@ import ExternalLink from './ExternalLink';
 import { graphql, useStaticQuery } from 'gatsby';
 import { css } from '@emotion/react';
 import useThemeTranslation from '../hooks/useThemeTranslation';
+import PrivacySwitch from '../icons/PrivacySwitch';
 import Trans from './Trans';
 import Link from './Link';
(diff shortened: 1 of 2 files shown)


tabathadelane (approved):
Looks good!

I did notice that when i click on it in mobile emulation, if i dismiss the error box, the icon stays highlighted. Not sure what that behavior will be like on actual mobile.

sunnyzanchi:
> Looks good!
> 
> I did notice that when i click on it in mobile emulation, if i dismiss the error box, the icon stays highlighted. Not sure what that behavior will be like on actual mobile.

hmm i wasn't getting this in Firefox. but i think it's okay if it's highlighted while it's focused on mobile, it should unhilight whenever the user taps somewhere else. i'm pretty sure that matches behavior i've seen before on buttons etc on mobile

PR #963, opened 2023-12-08
Add max height to images (#938)

tabathadelane:
* fix: Add max height to images

- also provides an 'unbound' class to unset this style

* fix: images overflowing main body container

---------


tabathadelane:
it says the build to develop is failing because of a package issue, but this isn't merging into develop so disregarding his check.

PR #972, opened 2024-01-05
feat: save selected tab in URL hash

sunnyzanchi:
when selecting a tab, the URL will update with the tab id. going to a URL with a hash will automatically select that tab

## testing

- go to theme demo
- try clicking different tabs
- ensure the URL updates with the tab id
- reload the page or copy-paste the URL to a different tab
- make sure the page switches to that tab on load

### DIFF
--- a/packages/gatsby-theme-newrelic/src/components/Tabs/Tabs.js
+++ b/packages/gatsby-theme-newrelic/src/components/Tabs/Tabs.js
@@ -58,6 +58,13 @@ const Tabs = ({ children, initialTab }) => {
         const index = pages.findIndex((page) => page.props.id === hash);
         if (index !== -1) {
           setTab(index);
+          const y =
+            tabsContainer.current.getBoundingClientRect().top +
+            window.pageYOffset +
+            // header height
+            72;
+
+          window.scrollTo({ top: y, behavior: 'smooth' });
           tabsContainer.current.scrollIntoView();
         }
       }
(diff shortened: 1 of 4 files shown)


tabathadelane:
1. Do we want the page to scroll to the tab group containing the tab? It's in the ticket AC, but I am wondering it that the expected behavior for something like this?
2. Mobile tabs only update the content and not the tab header in the dropdrown

sunnyzanchi:
> 1. Do we want the page to scroll to the tab group containing the tab? It's in the ticket AC because i think this was why we opted to only save one tab id in the URL for this reason, but I am wondering it that is actually the expected behavior for something like this?
> 
> 2. Mobile tabs only update the content and not the tab header in the dropdrown

1. hmm i think that makes sense. i'll update this
2. good catch! i'll push up a fix

packages/gatsby-theme-newrelic/src/components/Tabs/Tabs.js
```diff
@@ -30,6 +32,38 @@ const Tabs = ({ children, initialTab }) => {
...
+      // </Tabs>
+      // ```
+      // if not, something is wrong and we'll only break the page
+      // by trying further.
+      if (tabPages.type === Pages) {
+        const pages = tabPages.props.children;
+        const index = pages.findIndex((page) => page.props.id === hash);
+        if (index !== -1) {
+          setTab(index);
+          tabsContainer.current.scrollIntoView();
```
clarkmcadoo, line 61: Having a bit of an issue with the `scrollIntoView()` as we're not able to get around the header height here, might need the ole:
```
const y = element.getBoundingClientRect().top + window.pageYOffset + yOffset;

window.scrollTo({top: y, behavior: 'smooth'});
```

clarkmcadoo (approved):

PR #993, opened 2024-01-23
Create segment page view tracker

LizBaker:
This makes edits to our previous page view tracker which used tessen with some segment settings. Now we are making calls directly to segment and removing tessen 🎉


packages/gatsby-theme-newrelic/src/utils/page-tracking/segment.js
```diff
@@ -0,0 +1,67 @@
...
+    section && platform,
+    "You have enabled page view tracking, but do not have page view tracking configured. This route change has not been tracked. Please configure the 'segment.section' and 'segment.platform' options in gatsby-config.js"
+  );
+};
+
+const trackPageView = ({ config, env, location, prevLocation, analytics }) => {
+  const { section, platform } = config;
+  const getCookie = (key) => Cookies.get(key)?.replace(/%22/g, '') || null;
+  const customer_user_id = getCookie('ajs_user_id');
+  const anonymousId = analytics.instance?.user().anonymousId();
```
clarkmcadoo, line 18: We're using a cookie for this one: `getCookie('ajs_anonymous_id');` in the tessen sunset PR. Do they need to match or nah?

LizBaker: i'm not sure 😓 , this call will get the anonymous id for a user if it exists and otherwise create a new one

clarkmcadoo: oh that makes sense, this was set by segment anyway i believe

clarkmcadoo (dismissed):

packages/gatsby-theme-newrelic/src/utils/config/index.js
```diff
@@ -1,5 +1,6 @@
 module.exports = {
   getI18nConfig: require('./i18n'),
   getResolvedEnv: require('./resolvedEnv'),
+  getSegmentConfig: require('./segment'),
   getTessenConfig: require('./tessen'),
```
sunnyzanchi, line 5: should we remove `getTessenConfig`?

packages/gatsby-theme-newrelic/src/utils/page-tracking/segment.js
```diff
@@ -0,0 +1,67 @@
...
+  if (!canSendPageView(config)) {
+    return warnAboutNoop(section, platform);
+  }
+
+  analytics.page(`${platform} page_viewed`, {
+    event_type: 'page_view',
+    path: location.pathname,
+    referrer: prevLocation?.href,
+    section,
+    platform,
```
tabathadelane, line 29: It looks like we are using 'platform' to denote part of the event name as well as that `platform` field in the properties. `platform` on line 29 will always be 'docs_pages' for all off our teams sites (which is confusing), so that can be hard coded. We might rename this key in the gatsby-config where we are setting what site it's coming from.

tabathadelane (changes requested):

tabathadelane (approved):

PR #999, opened 2024-01-30
fix: add custom attributes to browser agent events

tabathadelane:
Add `customer_user_is` and `loggedIn` attributes to NR Browser Agent events

You can see the data being reported [here](https://staging.onenr.io/0gbRKxreOjE).
(_`loggedIn` NRQL query will always return `false` outside of `*.newrelic.com` or `localhost`_)


clarkmcadoo (approved):

sunnyzanchi (dismissed):
i appreciate the comment, v helpful

sunnyzanchi (approved):

PR #1052, opened 2024-05-07
fix: use push state vs navigation to keep page from scrolling

clarkmcadoo:
This might be a quick W


sunnyzanchi (approved):
nice! works great!

PR #1094, opened 2024-10-15
Update locales to include french

LizBaker:
🎉


packages/gatsby-theme-newrelic/src/utils/config/i18n.js
```diff
@@ -38,6 +38,13 @@ const LOCALE_CONFIGS = {
     hrefLang: 'pt',
     isDefault: false,
   },
+  fr: {
```
gmanandhar-nr, line 41: Do we need to add 'fr' [here](https://github.com/newrelic/gatsby-theme-newrelic/pull/1094/files#diff-48391d4c5a56cd1958c40473c4c3a68d123b5db455ea3e3c157a5187c2546f00R2) as well?

LizBaker: 💀 yep definitely. without it it won't pick up the french json we use for a lot of our components but i didn't test that because we don't have any json in french yet, whoops

PR #1095, opened 2024-10-15
search functionality

sunnyzanchi:
this PR adds functionality to the new search dropdown

## testing

- try searching for a few common search terms
- use arrow keys to navigate up and down
- try hitting enter with a recent query selected, a search selected, and with nothing selected
- try clicking results
- try resizing to mobile sizes
- hitting <kbd>/</kbd> focuses search from anywhere on the page

### DIFF
--- a/packages/gatsby-theme-newrelic/src/components/GlobalHeader.js
+++ b/packages/gatsby-theme-newrelic/src/components/GlobalHeader.js
@@ -1,9 +1,9 @@
-import React, { useState, useRef } from 'react';
+import React from 'react';
 import PropTypes from 'prop-types';
 import useMedia from 'use-media';
 import path from 'path';
 import { css } from '@emotion/react';
-import { graphql, useStaticQuery, Link } from 'gatsby';
+import { graphql, useStaticQuery } from 'gatsby';
 import { useLocation } from '@reach/router';
 
 import AnnouncementBanner from './AnnouncementBanner';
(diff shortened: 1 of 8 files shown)


packages/gatsby-theme-newrelic/src/components/GlobalHeader.js
```diff
@@ -537,10 +400,62 @@ const GlobalHeader = ({ className, activeSite, hideSearch = false }) => {
   );
 };
 
+export const NR_SITES = {
```
clarkmcadoo, line 403: Can we move this to the constants file at `packages/gatsby-theme-newrelic/src/utils/constants.js`?

packages/gatsby-theme-newrelic/src/components/GlobalHeader.js
```diff
@@ -537,10 +400,62 @@ const GlobalHeader = ({ className, activeSite, hideSearch = false }) => {
...
+  LEARN: 'LEARN',
+};
+
 GlobalHeader.propTypes = {
   className: PropTypes.string,
   activeSite: PropTypes.oneOf(Object.values(NR_SITES)),
-  hideSearch: PropTypes.bool,
+};
+
+const HEADER_LINKS = new Map();
```
clarkmcadoo, line 414: Could probably go into constants as well

packages/gatsby-theme-newrelic/src/components/GlobalSearch.js
```diff
@@ -0,0 +1,215 @@
...
+            width: 1rem;
+            height: 1rem;
+          }
+
+          input {
+            border: none;
+            height: 40px;
+          }
+
+          // TODO: use right bp here
```
clarkmcadoo, line 147: Seems like the right bp, is there anything that needs to get updated here?

packages/gatsby-theme-newrelic/src/components/SearchModal/useSearch.js
```diff
@@ -1,5 +1,5 @@
 import { useCallback, useState, useEffect, useReducer, useMemo } from 'react';
-import { useDebounce } from 'react-use';
+import { useThrottle } from 'react-use';
```
clarkmcadoo, line 2: I don't think we need this here anymore

packages/gatsby-theme-newrelic/src/components/GlobalSearch.js
sunnyzanchi: hmm well technically this breakpoint is specified in the theme config. but this is the right breakpoint from the designs and i don't think it would be reliable for this to be configurable. i'll remove this todo

PR #1166, opened 2026-04-29
fix: change privacy text in the footer

akashreddy-nr:


Nandu-pns (approved):

packages/gatsby-theme-newrelic/src/components/GlobalFooter.js
```diff
@@ -162,9 +164,8 @@ const GlobalFooter = ({ className }) => {
                   />
                 </Button>
               )}
-
-              <ExternalLink href="https://newrelic.com/termsandconditions/cookie-policy">
-                {t('footer.cookiePolicy', 'Cookie Policy')}
```
Nandu-pns, line 167: Since `Cookie Policy` has been removed from the footer, it looks the `cookiePolicy` translation key is now unused across all locale files. Worth removing these entries as cleanup ?

packages/gatsby-theme-newrelic/src/i18n/translations/en.json
```diff
@@ -52,8 +52,9 @@
     "careers": "Careers",
     "terms": "Terms of Service",
     "dmcaPolicy": "DMCA Policy",
-    "privacyNotice": "Privacy Notice",
+    "privacyNotice": "Privacy",
     "patentNotice": "Patent Notice",
+    "websiteTerms": "Website Terms",
```
Nandu-pns, line 57: The new `websiteTerms` key has been added to `en.json`, but is missing in other locale files. Do you think this need to be added in other files as well ?

PR #1169, opened 2026-06-25
feat(search): migrate from Swiftype to SearchGPT REST API

akashreddy-nr:
## Summary

Replaces the Swiftype search integration with New Relic's **SearchGPT REST API (v2)** across all theme-side search surfaces. Part of [NR-580466](https://new-relic.atlassian.net/browse/NR-580466) (FY27Q2: Docs Engineering — AI-Native Search).

REST was chosen over NerdGraph/GraphQL because its two-tier model fits our two UI surfaces and sidesteps the rate-limit problem on a public, single-key site:
- **`/v2/search/suggest`** — lexical typeahead, **not rate limited** → powers the as-you-type dropdown.
- **`/v2/search`** — hybrid, cursor-paginated, full bodies (cache-aware, rate-limited) → powers the results page.

## What changed

- **New `src/utils/searchGPT.js` client** — `search()`, `suggest()`, `fetchTags()`, `RateLimitError`. Base URL + api-key are env-driven (`GATSBY_SEARCHGPT_BASE_URL` / `GATSBY_SEARCHGPT_API_KEY`) so a backend proxy can slot in later with no code change.
- **Header dropdown** (`useSearch` + `GlobalSearch`) → `/v2/search/suggest`; "View more" grows the suggest `limit` to append results inline (same UX as before).
- **Public `search()` export** → re-points at `/v2/search` with a normalized shape `{ results, nextCursor, prevCursor, totalCount }`. **Breaking shape change** for consumers — see follow-ups.
- **404 page** suggestions → `suggest()`.
- **Build-time related-resources** → `/v2/search/suggest` (unthrottled, per SearchGPT docs guidance), excluded URLs filtered client-side.
- **New `src/utils/highlightText.js`** — client-side highlighting of query terms in titles (SearchGPT only highlights `summary`/`bodyHighlights`); HTML-escaped (XSS-safe).
- Removed unused `SWIFTYPE_ENGINE_KEY`; updated README (env vars + relatedResources docs).

## Testing

- `yarn test` — 90/90 passing; lint clean on all changed files.
- Client validated against the live API (cursor pagination, suggest-has-no-body, encoding).
- Locally smoke-tested via the demo (`/search-test` + `/search-results`) behind a dev proxy.

## Intentionally retained / deferred

- **`data-swiftype-index` attributes kept** — Swiftype is still live during cutover; removing them would pollute its index. Remove only after Swiftype is decommissioned.
[...]

PR #1173, opened 2026-07-29
feat(search): support a limit param on /v2/search (5 results per page)

akashreddy-nr:
## Summary

Adds a `limit` parameter to the SearchGPT `search()` client so the results page can request **5 results per page**, restoring parity with the current docs.newrelic.com page size.

Follow-up to #1169 (the Swiftype → SearchGPT REST migration, shipped in `9.16.0`). This is the one remaining piece from that work.

## Changes

- `utils/searchGPT.js` — `search()` accepts and forwards `limit` to `GET /v2/search`.
- `SearchModal/search.js` — public export wrapper threads `limit` through.
- Verified against `/v2/search`: `limit=5` returns 5 results/page and the cursor advances correctly (page 2 has zero overlap with page 1).

## Release

`feat:` commit → minor bump. On the next `develop → main` promotion, lerna will publish `9.17.0`. docs-website then bumps `@newrelic/gatsby-theme-newrelic` to consume it.

## Notes / follow-ups (not in this PR)

- **Numbered pagination** is not yet possible — `/v2/search` exposes only opaque cursors (no `offset`/`page`), so random page access requires an API change. Raised with the SearchGPT team.
- Other open items with the SearchGPT team: locale support, shared-key rate-limit behavior, exclude-from-indexing.


WriteMayur (approved):
looks good

PR #1176, opened 2026-07-31
fix(search): preserve locale when navigating to search-results

akashreddy-nr:
## Summary

Fixes a locale-drop bug in search navigation. On localized sites (e.g. `/fr/`), the header search typeahead correctly calls the SearchGPT `suggest` API with `language=fr`, but pressing **Enter** navigated to an absolute `/search-results/` path **with no locale prefix**. That dropped the `/fr/`, landed the user on the English results page, and — because the results page derives its language from `useLocale()` on the URL — scoped the actual `search` query to `language=en`. The wrong language was a downstream symptom; the real bug was the unlocalized navigation.

## Root cause

Three navigations built an absolute `/search-results/?query=…` string and passed it straight to `navigate()` / an `<a href>`, bypassing the theme's locale prefixing:

- `GlobalSearch.js` — header search `onSubmit` (both the fresh-query and selected-recent-query branches)
- `SearchDropdown.js` — the "Recent search terms" links
- `404.js` — the 404-page search input `onFocus`

## Fix

Route all three through the existing `localizePath({ path, locale })` helper (`utils/localization.js`), with `locale` from `useLocale()`. `localizePath` is a **no-op on the default English site** and prefixes the locale otherwise, so:

- `docs.newrelic.com/search-results/…` is unchanged.
- `docs.newrelic.com/fr/…` → `/fr/search-results/…`, preserving the locale and letting the results page resolve `language=fr`.

This matches how the theme `Link` component already localizes internal paths.

## Release

Conventional-commit `fix:` → **patch** bump (9.17.0 → **9.17.1**).

## Verification

- `eslint` passes on all three changed files.
- Behavior to confirm on a deploy preview: on `/fr/`, Enter from the header search (and clicking a recent term) navigates to `/fr/search-results/…` and the results call goes out with `language=fr`; on the default English site the URLs are unchanged.

### DIFF
--- a/packages/gatsby-theme-newrelic/src/components/GlobalSearch.js
+++ b/packages/gatsby-theme-newrelic/src/components/GlobalSearch.js
@@ -5,9 +5,11 @@ import { css } from '@emotion/react';
 import { useThrottle } from 'react-use';
 
 import useKeyPress from '../hooks/useKeyPress';
+import useLocale from '../hooks/useLocale';
 import useThemeTranslation from '../hooks/useThemeTranslation';
 import useScrollFreeze from '../hooks/useScrollFreeze';
 import { addPageAction } from '../utils/nrBrowserAgent';
+import { localizePath } from '../utils/localization';
 
 import useSearch from './SearchModal/useSearch';
 import SearchInput from './SearchInput';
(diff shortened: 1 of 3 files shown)


WriteMayur (approved):
lgtm

